PR #20, opened 2017-10-04
Adding student directory john_navitsky

johnnzz:
add student directory john_navitsky

### DIFF
--- a/students/john_navitsky/README.md
+++ b/students/john_navitsky/README.md
@@ -1,0 +1,2 @@
+Welcome to John Navitsky's (john@navitsky.org) directory.
+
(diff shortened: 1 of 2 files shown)


codedragon (approved):
Thanks!

students/john_navitsky/README.md
```diff
@@ -0,0 +1,2 @@
+Welcome to John Navitsky's (john@navitsky.org) directory.
+
```
codedragon, line 2: Fyi, since this is on github, you are essentially publishing your email address. You might not care (my email address is certainly lots of places on the web), but just wanted to make sure you were aware. Of course, since the purpose of git is to keep history, editing it out now will not completely erase it... But, it will bury it, which is probably sufficient, if you care. Spam filters are getting pretty good these days, anyway. ;-)

codedragon (approved):

students/john_navitsky/README.md
johnnzz: good point.  thx.

PR #30, opened 2017-10-06
Uploading break_me.py file.

brhelodrvr:


session01/break_me.py
```diff
@@ -0,0 +1,46 @@
...
+		print("Your type is incorrect:", e)
+
+# SyntaxError test:
+def syntax_test():
+	''' This tests SyntaxError '''
+	try:
+		date = eval('datetime(2009, 12a, 31)')
+	except SyntaxError as e:
+		print("Your syntax is wrong: ", e)
+
```
codedragon: I like how you got around that syntax errors are usually found on loading, rather than during runtime!

codedragon (changes requested):
Good job. Keep in mind that when you print the exception without raising it, you lose the stacktrace. There are ways around this, which we will talk about later, but one way is to log it when you catch it.
Notice that this file is not in your personal folder in the students folder. Please move the session01 directory you created to your folder and the do another pull.

brhelodrvr:
Moved session01/break_me.py into my brian_warn students directory.

session01/break_me.py
brhelodrvr: I just completed the https://uwpce-pythoncert.github.io/IntroPython-2017/modules/GitWorkflow.html#git-workflow process to make this change.  Pls let me know if I succeeded.

PythonCHB: that was clever -- but do remember that using "eval" in production code is discouraged, mostly for security reasons.

works well for this though.

PythonCHB:
file is still showing as in the main dir.

you need to do a couple things:

1) on your laptop, move the file IN GIT:

```
git mv session01/break_me.py students/brhelodrv/break_me.py
git commit -a
```
then push the changes to gitHub:

```
git push
```
you may also want to pull in the the latest version of the full class repo:

```
git pull upstream master
```
If this is all confusing, we'll go over it in class Tuesday

brhelodrvr:
I’d appreciate hearing more about eval — reasons not to use it and preferred/suggested alternatives — to it either here or in class tomorrow.  I don’t recall if eval was discussed during the Foundations course.  I found references to eval when looking for ways to raise SyntaxError.

brhelodrvr:
Looks like I’ll need some level of hand holding to get past this issue.  I appear to be getting tripped up by the brhelodrvr user that I created on GitHub some time ago for some Udemy or equivalent course I was doing at that point.  From the output below, I believe that I need to set up a new account for brian_warn and then shift my files to it.  Probably best at this point to make the switch in-person since git isn’t intuitive (yet) for me.

raphael:brian_warn brwarn$ git pull upstream master
From https://github.com/brhelodrvr/IntroPython-2017
 * branch            master     -> FETCH_HEAD
Already up-to-date.
[master>]
raphael:brian_warn brwarn$ pwd
/Users/brwarn/Python2017/IntroPython-2017/students/brian_warn
[master>]
raphael:brian_warn brwarn$ ls ..
EricAdams	README.md	brian_warn	gene		scott
Larry		alinafe		davidkan	kegan		shibata
[master>]
raphael:brian_warn brwarn$ ls 
readme.txt	session01
[master>]
raphael:brian_warn brwarn$ ls session01/
break_me.py
[master>]
raphael:brian_warn brwarn$ git mv session10/break_me.py students/brh
fatal: bad source, source=students/brian_warn/session10/break_me.py, destination=students/brian_warn/students/brh
[master>]
raphael:brian_warn brwarn$

PythonCHB:
In short: "eval" evaluates code at run-time.

If that code is coming from an untrusted source, it could be very dangerous.

If it is a trusted source, it is still doing the whole parse, compile and run each time through the code, so not very efficient.

As for alternatives, it depends on your use case, but Python allows you to manipulate objects pretty directly, so there is very rarely a use-case for eval. But we'll be getting to that (metaprogramming) in the second quarter.

As for how to get a syntax error -- just put it in the code -- the code won't run, as the syntax error will prevent it from being compiled to byte code -- but that was kinda the point of the exercise :-)

You'll be getting plenty of them as you work on code -- at least I do!

Ask in class if you want to know more.

PythonCHB:
Ah! I see:

Yes, this is happening as the "brhelodrvr" gitHub user. If you have more than one gitHub ID, things can get confusing!

though that's not a git issue per se -- it's a gitHub one. In short, make sure you are logged in as the user you expect to be whenever you do anything.

We can clean this up in class.

brhelodrvr:
Chris,
The eval explanation makes sense — thank you.  

SyntaxError is a rather obvious one in retrospect.  It did however enable you to point out to me the security aspects that exist with its use.

Thanks again,
Brian

brhelodrvr:
Very good!

PR #32, opened 2017-10-08
Hey, Maria. Hopefully this would be helpful.

zandra2:


students/zandra_eng/session01/break_me.py
```diff
@@ -0,0 +1,31 @@
...
+    a = 5
+    a += "aa"
+
+def syntax_error():
+    a 5 .. ;;;;
+
+def attribute_error():
+    a = "string"
+    a.whatever()
+
```
codedragon, line 31: Looks great. As is, will cause the syntax error. Syntax errors are mostly found during load, not run, so you have to comment out the function itself to stop the error. The rest of the errors are caused by actually running the respective functions.

codedragon (approved):

PR #37, opened 2017-10-11
adding break_me content

briannagata:


codedragon (commented):
Nice tests! I'm afraid we weren't clear enough in the objective, however. We were looking for you to figure out what sort of things causes these errors to be raised, rather than raising them explicitly.

students/briannagata/session01/break_me.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+def syntax_error():
+    """ Raises exception SyntaxError """
+    return eval('')
+
+
+def attribute_error():
+    """ Raises exception AttributeError """
+    return list('attribute').split()
```
PythonCHB, line 22: you got it!

students/briannagata/session01/test_break_me.py
```diff
@@ -0,0 +1,26 @@
+# break_me test
```
PythonCHB, line 1: Ahead of the game with this one -- but really nice.

students/briannagata/session02/fizzbuzz.py
```diff
@@ -0,0 +1,18 @@
...
+# Python 3.6
+
+
+def fizzbuzz():
+    """
+    From an inclusive range from 1 - 100, print whether the number is a
+    multiple of 3, 5, or both and print the associated Fizz/Buzz words.
+    """
+    for cycle in range(1, 101):
+        if cycle % 3 is 0 and cycle % 5 is 0:
```
PythonCHB, line 11: You do not wnat to use "is" for this test.

"is" tests whether the objects are actually the same object --not whether they have the same value. AS you can easily have multiple objects that happen to have the same value, "is" will fail in the general case.

This works because cpython has an optimization called "interning" -- since small integers are used so often, the interpreter keeps a pool if them around to re-use, rather than creating multiple ints with the same value.

so "is" will work as a test for small integers, but not large ones:

```
In [62]: x = 345678

In [63]: y = 345678

In [64]: x is y
Out[64]: False
```

You really want if cycle % 3 == 0 and cycle % 5 == 0:

students/briannagata/session02/grid_printer.py
```diff
@@ -0,0 +1,27 @@
+# Grid printer exercise
+# Python 3.6
+
+
+def print_grid(number):
+    """ Print a grid expanded by number provided """
+    half = int(number / 2)
```
PythonCHB, line 7: you can use the // operator to get "floor division" -- always an integer.

students/briannagata/session02/grid_printer.py
```diff
@@ -0,0 +1,27 @@
...
+def print_grid2(cell_size, scale):
+    """ Print a grid with specific cell count and expanded by scale """
+    top = ''.join(['+', ' - ' * scale])
+    sides = ''.join(['|', '   ' * scale])
+
+    for count in range(cell_size):
+        print(top * cell_size, '+')
+        for size in range(scale):
+            print(sides * cell_size, '|')
+    print(top * cell_size, '+')
```
PythonCHB, line 27: this all looks good.

students/briannagata/session02/series.py
```diff
@@ -0,0 +1,35 @@
...
+    """
+    Returns a generalized result with option to define the first
+    two values. Default values will produce fibonacci sequence.
+    """
+    if n is 0:
+        return first
+    elif n is 1:
+        return second
+
+    return sum_series(n-2, first, second) + sum_series(n-1, first, second)
```
PythonCHB, line 35: perfect!

students/briannagata/session02/test_series.py
```diff
@@ -0,0 +1,28 @@
+# Test for series
+# Python 3.6
+
+
+import pytest
+import series
+
+
+@pytest.mark.parametrize('n, expected', [(0, 0),
```
PythonCHB, line 9: VERY nice! The rest of us won't get to this for weeks .....

PythonCHB (approved):
All looks good.

I love the tests -- we'll be getting to that in a couple weeks, but great to start now.

PR #38, opened 2017-10-11
break_me and session02 submissions

mattmaeda:


students/mattmaeda/session01/break_me.py
```diff
@@ -0,0 +1,43 @@
...
+    except Exception as e:
+        print("#### Name Error ####")
+        print(str(e) + "\n")
+
+    try:
+        raise_type_error()
+    except Exception as e:
+        print("#### Type Error ####")
+        print(str(e) + "\n")
+
```
codedragon, line 31: So, this is a clever way to avoid commenting out the functions, but doing it this way masks the error that you are getting. If you run your raise_type_error function outside of your try, except block you will notice that you are actually getting a name error, because a has not assigned to anything. So, to use your method to avoid commenting out things, but still make sure you are getting the correct error is to use `except NameError as e:` (or whichever error you are testing for). Generally speaking, you never want to catch *all* exceptions, but only the ones that you know how to deal with.

students/mattmaeda/session02/grid_printer.py
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/python
+
+def print_line(intersection_char, fill_char):
+    fill = "{} ".format(fill_char)
+    print(intersection_char, end=' ')
+    print(fill * 4, end='')
+    print(intersection_char, end=' ')
+    print(fill * 4, end='')
+    print(intersection_char)
+
```
codedragon: Good solution. Notice you are repeating some lines. How might you change your code to get rid of the repetition?

students/morgan/session01/break_me.py
```diff
@@ -0,0 +1,15 @@
+
```
codedragon, line 1: I'm not sure why this file is here. It is in someone else's directory, and it doesn't look different from the one that is in the repo. I assume you must have done a git pull somewhat recently. I think something must have gone weird with a merge. I'm going to leave it as is, and see how it goes...

students/mattmaeda/session02/series.py
```diff
@@ -0,0 +1,38 @@
...
+    assert fibonacci(3) == 1    # fibonacci(3) == 1
+    assert fibonacci(5) == 3    # fibonacci(5) == 3
+    assert fibonacci(10) == 34  # fibonacci(10) == 34
+
+    assert lucas(1) == 2        # lucas(1) == 2
+    assert lucas(2) == 1        # lucas(2) == 1
+    assert lucas(3) == 3        # lucas(3) == 3
+    assert lucas(5) == 7        # lucas(5) == 7
+    assert lucas(10) == 76      # lucas(10) == 76
+
```
codedragon, line 37: For similar reasons your lucas numbers are not quite correct either.

codedragon (approved):

PR #39, opened 2017-10-11
Adding break_me.py homework

dkkan:
Adding break_me.py homework for review.


session01/break_me.py
```diff
@@ -0,0 +1,94 @@
+# Description: In the break_me.py file write four simple Python functions:
+# Each function, when called, should cause an exception to happen
+# Each function should result in one of the four most common exceptions you’ll find.
+# for review: NameError, TypeError, SyntaxError, AttributeError
+# Comment: Execute this script in Python3.6
+# Last modified: 10/07/2017 - davidkan@
+
+
+class PracticeExceptions:
```
PythonCHB, line 9: we haven't gotten to classes yet. Nothing wrong with using them when appropriate, but this is not a good place for a class.

It's not considered "python" style to use a class as a way to hold a bunch of staticmethods -- no point in that.

PythonCHB (approved):
We also weren't expecting you to catch the exceptions at this point -- but nice to do that.

PR #41, opened 2017-10-11
it should work now

rajarameshyv:


students/ramesh/print_grid.py
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env python
+
+"""
+print a grid
+
+"""
+def plus_line(x):
```
PythonCHB, line 7: nice job breaking it into functions.

However, it would be cleaner (and more extendible) to have the functions return the string to print -- and then do the actual printing in the calling function. For example:

```
def line_line(x):
    return '|' + x * '   ' + '|' + x * '   ' + '|'
```
and 

```
    for i in range(x):
        print(line_line(x))
```
that would let you clean it up a bit more with:

```
print line_line(x) * x
```
rather than the for loop.

students/ramesh/ramesh.out
```diff
@@ -0,0 +1 @@
+new code
```
PythonCHB, line 1: what's this file? do you want it in git?

beware of `git add *` -- it can very easily add support files you don't want in the repo

PR #43, opened 2017-10-11
session02 in class activities added

kathrynegan:


students/kegan/session01/2017 10 03 Session 1.txt
```diff
@@ -0,0 +1,2 @@
+questions:
+1) any advantage to using del versus just reassigning or ignoring the variable?
```
PythonCHB, line 2: del is there if you need it, but you are quite right — it’s actually pretty rare to need it.

Most often you can re-use the name as you say, or simple ignore it — it will usually go out of scope soon anyway. — at the end of a function, for instance.”

students/kegan/session01/break_me.py
```diff
@@ -0,0 +1,49 @@
...
+author@kathrynegan
+
+Causes and catches NameError, AttributeError,
+and TypeError in code and prints message if
+error was successfully generated.
+SyntaxError code included and commented out.
+"""
+
+
+def cause_name_error():
```
PythonCHB, line 11: Perfect -/ and you are ahead with the try-except!

students/kegan/session02/print_grid.py
```diff
@@ -0,0 +1,54 @@
...
+|         |
+|         |
+|         |
+* - - - - *
+"""
+import argparse
+
+
+def main():
+    parser = argparse.ArgumentParser()
```
PythonCHB, line 17: Really going for it here! Nice!

students/kegan/session02/print_grid.py
```diff
@@ -0,0 +1,54 @@
...
+    args = parser.parse_args()
+    # print row by column grid
+    for i in range(args.row):
+        print_horizontal(args.column)
+        for i in range(4):
+            print_vertical(args.column)
+    print_horizontal(args.column)
+
+
+def print_horizontal(column):
```
PythonCHB, line 33: Nice job breaking it out into functions.

But if you going to do that -/ a bit better to have the functions return a string, and do the printing when you call it. More flexible and extendable that way — say you wanted to write it to a file?

students/kegan/session02/series.py
```diff
@@ -0,0 +1,84 @@
...
+        int : current value
+    """
+    if n == 1:
+        return 2
+    if n == 2:
+        return 1
+    return lucas(n - 2) + lucas(n - 1)
+
+
+def sum_series(n, base0, base1):
```
PythonCHB, line 38: Perfect!

students/kegan/session02/series.py
```diff
@@ -0,0 +1,84 @@
...
+    if n == 1:
+        return base0
+    if n == 2:
+        return base1
+    return (
+        sum_series(n - 2, base0, base1) +
+        sum_series(n - 1, base0, base1))
+
+
+def test(module, solutions, **kwargs):
```
PythonCHB, line 57: Very slick! 

Are you sure you haven’t taken this class before?

We willl be introducing “proper” unit testing in a few weeks, then you won’t need to write utilities like this :-)

students/kegan/session02/series.py
```diff
@@ -0,0 +1,84 @@
...
+    """ Tests given module against a solution.
```
PythonCHB, line 58: I would use the name “function “ here. “Module” has a specific ( and different) meaning in Python.

PythonCHB (approved):

kathrynegan:
I have not taken the class before :) but I have some experience in Python
2.7. Looking to really nail down some skills in basic software development
and unit testing, and get feedback on how to make my code as pythonic as
possible. Thanks for the comments!

PR #45, opened 2017-10-12
Solutions to series.py, fizzbuzz, and print grid

Eowyn42:
I've explored a few different ways to solve the print grid problem. Ultimately I realized there are symmetries - each line is constructed the exact same way except some of them use plus and minus characters and some use dash and space characters. 

Once I had solved that problem, I returned to the first couple of problems using this approach, except now the first argument was gone / fixed equal to 2, and the second argument is something like (n-1)/2. ie n =3 is the same shape as ndim = 2, ndash = 1 ; n = 15 is the same shape as ndim = 2, ndash = 7. But I'm having problems with the n = 11 case because it looks like the same shape as ndim = 2 and ndash = 4 but my solution produces ndim = 2, ndash = 5. I cribbed/tested/used a little rounding function from the internet, but neither rounding nor straight // truncation produced the desired result. I'm not sure if there's something I'm missing in the Part 2 part of the problem. . . my solution for Part 3 can solve Part 1 and MOST of Part 2, just not the n = 11 case.

### DIFF
--- a/students/eowyn/session02/print_grid.py
+++ b/students/eowyn/session02/print_grid.py
@@ -52,6 +52,7 @@ def print_grid(ndim,ndash):
     print(make_row(ndim,ndash,'+','-'))
 
 ### Now try the simple case where ndim is always 2
+### This is not producing the expected result for n = 11
 
 def roundno(no):
     ''' round to nearest whole integer '''
--- a/students/eowyn/session02/print_grid.py
+++ b/students/eowyn/session02/print_grid.py
@@ -1,0 +1,74 @@
+#!/usr/bin/env python
+
+''' Construct grid that is composed of ndim boxes in 
+    either direction, and each box is composed of ndash 
+    lines both horizontally and vertically '''
+
+### First method uses global vars and functions for lines and verts
+### then a caller function to print it all
+
+# define global varaibles
+plus = "+"
+minus = "-"
+space = " "
+dash = '|'
+
+def make_line(ndim,ndash):
+    line = ''
+    for n in range(ndim):
+        line += plus + ndash*minus
+    line += plus
+    return line
+
+def make_vert(ndim,ndash):
+    vert = ''
+    for n in range(ndim):
+        vert += dash + ndash*space
+    vert += dash
+    return vert
+
+def print_grid1(ndim,ndash):
+    for i in range(ndim):
+        print(make_line(ndim,ndash))
+        for j in range(ndash):
+            print(make_vert(ndim,ndash))
+    print(make_line(ndim,ndash))
+
+### Second method uses just one general construction function
+### and a caller function to print it all
+
+def make_row(ndim,ndash,char1,char2):
+    row = ''
+    for n in range(ndim):
+        row += char1 + ndash*char2
+    row += char1
+    return row
+
+def print_grid(ndim,ndash):
+    for i in range(ndim):
+        print(make_row(ndim,ndash,'+','-'))
+        for j in range(ndash):
+            print(make_row(ndim,ndash,'|',' '))
+    print(make_row(ndim,ndash,'+','-'))
+
+### Now try the simple case where ndim is always 2
+
+def roundno(no):
+    ''' round to nearest whole integer '''
+    return int(no//1 + ((no%1)/0.5)//1)
+
+def print_grid0(n):
+    ## handle odd n by rounding to nearest int
+    (ndim,ndash) = (2,roundno((n-1)/2))
+    print_grid(ndim,ndash)
+
+
+if __name__=="__main__":
+    print_grid(3,4)
+    print_grid(5,3)
+    print_grid(2,1)
+    print_grid(2,7)
+    print_grid(2,4)
+    print_grid0(3)
+    print_grid0(15)
+    print_grid0(11)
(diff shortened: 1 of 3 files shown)


PythonCHB:
Wow! nice thorough job! 

I"ll go look at the code now....

students/eowyn/session02/print_grid.py
```diff
@@ -0,0 +1,75 @@
+#!/usr/bin/env python
+
+''' Construct grid that is composed of ndim boxes in 
+    either direction, and each box is composed of ndash 
+    lines both horizontally and vertically '''
+
+### First method uses global vars and functions for lines and verts
+### then a caller function to print it all
+
+# define global varaibles
```
PythonCHB, line 10: In general, global variables are considered "bad form" -- but this is a fine use case --essentially constants.

It is considerd good style to use ALL CAPS for contants, so you may want to use names like:

PLUS = "+"
MINUS = "-"

and so on. But no biggie.

students/eowyn/session02/print_grid.py
```diff
@@ -0,0 +1,75 @@
...
+plus = "+"
+minus = "-"
+space = " "
+dash = '|'
+
+def make_line(ndim,ndash):
```
PythonCHB, line 16: another style point -- better to put a space after a comma: make_line(ndim, ndash).

students/eowyn/session02/print_grid.py
```diff
@@ -0,0 +1,75 @@
...
+    line = ''
+    for n in range(ndim):
```
PythonCHB, line 18: A loop is fine, but you could get the same effect with multiplication:

line = ndim * (plus + ndash*minus)

students/eowyn/session02/print_grid.py
```diff
@@ -0,0 +1,75 @@
...
+
+def print_grid1(ndim,ndash):
+    for i in range(ndim):
+        print(make_line(ndim,ndash))
+        for j in range(ndash):
+            print(make_vert(ndim,ndash))
+    print(make_line(ndim,ndash))
+
+### Second method uses just one general construction function
+### and a caller function to print it all
```
PythonCHB, line 38: nice! much better to put all the printing in one place.

students/eowyn/session02/print_grid.py
```diff
@@ -0,0 +1,75 @@
...
+    for i in range(ndim):
+        print(make_row(ndim,ndash,'+','-'))
+        for j in range(ndash):
+            print(make_row(ndim,ndash,'|',' '))
+    print(make_row(ndim,ndash,'+','-'))
+
+### Now try the simple case where ndim is always 2
+### This is not producing the expected result for n = 11
+
+def roundno(no):
```
PythonCHB, line 57: python does have a "round" function....

round(n)

or 

round (n/2) ....

students/eowyn/session02/series.py
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env python
+"""Return fibbonaci,lucas, and generic summed series series"""
+
+def fib(n):
+    """ Print the nth element of the fibbonaci sequence"""
+    return sum_series(n)
+
+def lucas(n):
```
PythonCHB, line 8: nice! once you've written the generic one -- use it to provide the others.

students/eowyn/session02/series.py
```diff
@@ -0,0 +1,37 @@
...
+    """ Print the nth element of the lucas sequence"""
+    return sum_series(n,2,1)
+
+def sum_series(n,n0=0,n1=1):
```
PythonCHB, line 12: perfect! and nice tests.

students/eowyn/session02/print_grid.py
```diff
@@ -0,0 +1,75 @@
+#!/usr/bin/env python
+
+''' Construct grid that is composed of ndim boxes in
```
PythonCHB, line 3: nice docstring -- but better style to put the quotes on their own line.

```
"""
Construct grid that is composed of ndim boxes in 
either direction, and each box is composed of ndash 
lines both horizontally and vertically
 """
```

students/eowyn/session02/print_grid.py
```diff
@@ -0,0 +1,75 @@
...
+
+def print_grid(ndim,ndash):
+    for i in range(ndim):
+        print(make_row(ndim,ndash,'+','-'))
+        for j in range(ndash):
+            print(make_row(ndim,ndash,'|',' '))
+    print(make_row(ndim,ndash,'+','-'))
+
+### Now try the simple case where ndim is always 2
+### This is not producing the expected result for n = 11
```
PythonCHB, line 55: it seems to work fine for me for n == 11 ? what is the issue?

PythonCHB (approved):
Very nice job!

Eowyn42:
Hi Chris,

Thank you for the code review. The n=11 issue is that the image on the
problem page (the .png with the white background, attached) has 4 dash
lines per segment whereas my solution has 5 (.png with the black
background, attached). And as they say, "five is right out!"

Thank you,
Eowyn

PythonCHB:
Ah -- I think that's 'cause the problem is not precisely specified -- there are multiple options for rounding, etc.

I think your solution is fine :-)

PR #46, opened 2017-10-13
Add fib_hiroTakechi.py

HiroTestStudent:
Hello,

I worked on the fibonacci codes after class and added three codes to the codes that we worked on in-class on Tuesday.

Thanks,
Hiro

### DIFF
--- a/students/hirot_directory/fib_hiroTakechi.py
+++ b/students/hirot_directory/fib_hiroTakechi.py
@@ -1,0 +1,53 @@
+
+'''
+Fibonacchi recursion Python code
+
+'''
+
+# in-class:
+
+def fib(n):
+
+	if n == 0:
+		return 0
+	if n == 1:
+		return 1
+	return fib(n-1) + fib(n-2)
+
+if __name__ == '__main__':
+	assert fib(0) == 0
+	assert fib(1) == 1
+	assert fib(2) == 1
+	assert fib(3) == 2
+	print(fib(10))
+
+# the codes that I worked on after class: 
+
+## code 1:
+def fibonacci(n, a=1, b=0):
+    return b if n < 1 else fibonacci(n - 1, a + b, a)
+
+print(fibonacci(7))
+
+
+## code 2:
+class Program():
+
+    def Fib(fib):
+        if fib == 0:
+            return 0
+        elif fib == 1:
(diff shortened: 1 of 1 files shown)


students/hirot_directory/fib_hiroTakechi.py
```diff
@@ -0,0 +1,53 @@
...
+	assert fib(0) == 0
+	assert fib(1) == 1
+	assert fib(2) == 1
+	assert fib(3) == 2
+	print(fib(10))
+
+# the codes that I worked on after class: 
+
+## code 1:
+def fibonacci(n, a=1, b=0):
```
PythonCHB, line 27: this is VERY slick! I've never seen this solution, but it's efficient and clean. Though kinda hard to follow the logic.

students/hirot_directory/fib_hiroTakechi.py
```diff
@@ -0,0 +1,53 @@
...
+    return b if n < 1 else fibonacci(n - 1, a + b, a)
+
+print(fibonacci(7))
+
+
+## code 2:
+class Program():
```
PythonCHB, line 34: we'll get to classes in a couple weeks, but this is not an appropriate use case.

And it doesn't work -- the Fib method is calling a global Fib()function -- or it will fail if there isn't one.

PythonCHB (approved):

students/hirot_directory/fib_hiroTakechi.py
HiroTestStudent: Hi Chris,

Thank you for the feedback. I just attended Maria's office hours and she also said we would cover the class and OOP in a few weeks. I just wanted to try it since Foundations of Python Programming class required to use their OOP template..

Thanks,
Hiro

students/hirot_directory/fib_hiroTakechi.py
HiroTestStudent: Hi Chris,

I think this is from the recursion exercise (or homework) of Foundations of Algorithm and Data Structure, if I remember correctly. 

Thanks,
Hiro

PythonCHB: you could have claimed credit :-)

PR #48, opened 2017-10-14
fizz_buzz_jake was completed during Week 2 Session.

jolsby:
Here is the fizz_buzz_jake file I built during class two of python.


students/py_directory_jake/session02/fizz_buzz_jake.py
```diff
@@ -0,0 +1,13 @@
+
```
PythonCHB, line 1: perfectly reasonable solution.

PR #49, opened 2017-10-14
This is fizzbuzz code.

HiroTestStudent:
Hi Maria,

This is my practice code on fizzbuzz assignment. I am still not clear of the fizzbuzz code that we worked in the class. (the code that doesn't require to code "print ("fizzbuzz"). 

Thanks,
Hiro

### DIFF
--- a/students/hirot_directory/fizz.py
+++ b/students/hirot_directory/fizz.py
@@ -1,0 +1,51 @@
+
+'''
+Dev: Hiro Takechi
+Subject: FizzBuzz
+'''
+
+'''
+def print_fizz():
+	print("Fizz")
+
+def print_buzz():
+	print("Buzz")
+
+def print_fizzbuzz():
+	print("FizzBuzz")
+'''
+
+def fizz():
+
+	print("approach #1")
+
+	for n in range(1,101):
+
+		if (n % 3) == 0 and (n % 5) == 0:
+			print(n, "FizzBuzz")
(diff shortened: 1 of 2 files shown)


students/hirot_directory/fizz.py
```diff
@@ -0,0 +1,51 @@
...
+'''
+
+def fizz():
+
+	print("approach #1")
+
+	for n in range(1,101):
+
+		if (n % 3) == 0 and (n % 5) == 0:
+			print(n, "FizzBuzz")
```
PythonCHB, line 25: you are supposed it print "FizzBuzz" instead of the number, not in addition to the number.

but the logic is fine.

PythonCHB:
Hiro,

The idea is that "FizzBuzz" is "Fizz" + "Buzz". And we want it when the number is a multiple of both 3 and five.

so rather than a separate print for FizzBuzz, you can do the "Fizz" when it's 3, and "Buzz" when it's 5, and then you'll get both when it's both.

```
msg = ""
if not n % 3:
    msg += "Fizz"
if not n % 5:
    msg += "Buzz" # here you will now have "FizzBuzz" if it was a multiple of 3, but only Buzz if not.
if not msg:
    msg += str(n)
print msg
```

PythonCHB (changes requested):
Why don't you remove the printing of the number when it's 3 or 5, and we'll call it good.

PR #50, opened 2017-10-14
Session02 .py files

ea8606:
This pull is for the session 2 .py files we worked on in class and at home


session01/break_me.py
```diff
@@ -1,3 +1,38 @@
+<<<<<<< HEAD
```
PythonCHB, line 1: This is cruft left over from a "merge conflict: form git.

When git tries to merge two versions, and it can't tell which one you want, it puts both in, with this stuff marking it.

You need to go in by hand and save the version of the code you want, and remove the merge markers.

Then you can commit that change, and all will be well.

students/EricAdams/session02/print_grid_part3.py
```diff
@@ -0,0 +1,30 @@
...
+
+def print_grid(cell_size,rows):
+    for i in range(rows):                   # rows = how many rows
+        print_plus_n_pipe(cell_size, rows)
+    print_line_plus(cell_size, rows)
+
+print_grid(4, 2)
+print() 
+print_grid(12, 3)   
+
```
PythonCHB, line 30: this all looks good --nice work.

students/EricAdams/session02/series.py
```diff
@@ -0,0 +1,34 @@
...
+        return 0
+    if n == 1:
+        return 1
+    return fibonacci(n-2) + fibonacci(n - 1)
+    
+
+def fib_series(n):
+    """ Print the nth value of the fibonacci series."""
+    print ('fibonacci(',n,') = ', fibonacci(n))
+    return 0
```
PythonCHB, line 13: Python will return None if you don't return anything -- and this is what you want if teh function doesn't return anything useful.

students/EricAdams/session02/series.py
```diff
@@ -0,0 +1,34 @@
...
+def lucas(n):
+    """ Return the lucas series nth value."""
+    if n == 0:
+        return 2
+    if n == 1:
+        return 1
+    return lucas(n-2) + lucas(n-1)
+
+def lucas_series(n):
+    """ Print the nth value of the fibonacci series."""
```
PythonCHB, line 24: don't forget to update your docstrings!

but great that you have them.

students/EricAdams/session02/sum_series.py
```diff
@@ -0,0 +1,52 @@
...
+        if n == 0:
+            return 2
+        elif n == 1:
+            return 1
+        else:
+            lucas_sum = sum_series(n-2, x=2, y=1) + sum_series(n-1, x=2,y=1)
+            return lucas_sum
+    else:
+        # print("Series not implemented")
+        return "Series not implemented"
```
PythonCHB: The idea here is that it could produce ANY series of this family, with any two starting values.

So it should be generic, rather than checking for those particular starting values.

and you can use the starting values as the termination criteria .. something like:

```
if n ==0:
    return x
```

etc....

PythonCHB (changes requested):
Why don't you fix the sum_series and we can merge.

-CHB

ea8606:
changed sum_series to be more general

PythonCHB:
This looks good.

And even if you don't want our review, it's still a good idea to keep the PR open and then we can merge.

That way, the next time around we know what's new and what's not.

-CHB

ea8606:
ya sorry about that.... still getting used to GitHub


________________________________

PR #51, opened 2017-10-15
mailroom and rot13

guillaumethomas:
Per your comments, I have written a less high level version of rot13.py

### DIFF
--- a/students/guillaume/session03/mailroom.py
+++ b/students/guillaume/session03/mailroom.py
@@ -1,0 +1,44 @@
+#!/usr/bin/env python3
+def read_data(textfile):
+    '''
+    '''
+    file = open(textfile, 'r')
+    donors = list()
+    for i in list(file.readlines()):
+        donors.append((i.rstrip()).split(','))
+        for i, s in enumerate(donors[-1]):
+            donors[-1][i] = s.lstrip()
+            if donors[-1][i].isdecimal():
+                print(s)
+                donors[-1][i] = float(s)
+    file.close()
+    return donors
+
+
+def input_sc():
+    '''
+    '''
+    welcome = """
+            local charity donor database\n
+
+                Please make a choice:
+
+                  1 - Send a thank you
+                  2 - Create a report
+                  3 - Quit\n
+                """
+    print(welcome)
+    bool = True
+    while bool:
+        selection = input('Please type a number between 1 & 3:\n')
+        if selection.isdecimal():
+            selection = int(selection)
+            if selection > 0 and selection < 4:
+                bool = False
+
+
+if __name__ == '__main__':
+    '''
+    '''
+    print(read_data('donors.txt'))
+    input_sc()
--- a/students/guillaume/session03/mailroom.py
+++ b/students/guillaume/session03/mailroom.py
@@ -1,6 +1,9 @@
 #!/usr/bin/env python3
 def read_data(textfile):
     '''
+    Read data from a textfile about the donor and the
+    amount of donations
+    Pattern is as follow:
     '''
     file = open(textfile, 'r')
     donors = list()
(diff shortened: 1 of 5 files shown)


students/guillaume/session03/mailroom.py
```diff
@@ -0,0 +1,142 @@
...
+    file = open(textfile, 'w')
+    for donor in donors:
+        donor_str = ', '.join(map(str, donor)) + '\n'
+        file.write(donor_str)
+    file.close()
+
+
+def add_donation(ind, donors):
+    '''
+    '''
```
PythonCHB, line 30: no need for the quotes if you aren't going to put anything in them!

students/guillaume/session03/mailroom.py
```diff
@@ -0,0 +1,142 @@
...
+    boolean = True
+    while boolean:
+        donation = input('\nPlease add a donation amount?\n')
+        if donation.isdecimal():
+            # need to work on test for float donation
```
PythonCHB, line 35: we'll get to that in a couple weeks!

students/guillaume/session03/mailroom.py
```diff
@@ -0,0 +1,142 @@
...
+    for donor in donors:
+        donor_str = ', '.join(map(str, donor)) + '\n'
+        file.write(donor_str)
+    file.close()
+
+
+def add_donation(ind, donors):
+    '''
+    '''
+    boolean = True
```
PythonCHB, line 31: better to use a meaningful name, rather than "boolean".

students/guillaume/session03/mailroom.py
```diff
@@ -0,0 +1,142 @@
...
+    while boolean:
```
PythonCHB, line 32: The other option is to use:

```
while True:
    ...
    if it_is_time_to_end_the_loop:
        break
```

students/guillaume/session03/mailroom.py
```diff
@@ -0,0 +1,142 @@
...
+
+            I would like to thank you personnaly for the generous
+            donation of ${} that you have made recently.
+            Your committment to our cause is very valuable to us.
+
+            Best regards,
+
+                                    John Adams
+                                    CEE of charity {}
+            '''.format(donor, donation, charity)
```
PythonCHB, line 56: very nice!

students/guillaume/session03/mailroom.py
```diff
@@ -0,0 +1,142 @@
...
+        else:
+            print('Make sure you enter Last & first name\n')
+
+    write_data(textfile, donors)
+    email = email_thank_you(name, donation)
+    return email
+
+
+def report(textfile):
+    donors = read_data(textfile)
```
PythonCHB, line 91: probably better to load the data into a data structure in memory, and then work with it in all the functions.

students/guillaume/session03/slicing_lab.py
```diff
@@ -31,11 +31,7 @@ def reverse_slic(seq):
     '''
     return a sequence reversed (just with slicing)
     '''
-    l = len(seq)
-    seq_ret = seq[l - 1:l]
-    for i in range(l - 1, -1, -1):
-        seq_ret = seq_ret + seq[i - 1:i]
-    return seq_ret
+    return seq[::-1]
```
PythonCHB, line 34: you got it!

students/guillaume/session03/slicing_lab.py
```diff
@@ -46,15 +42,19 @@ def third(seq):
...
     return None
 
+
 if __name__ == '__main__':
 
-    functions = [exch_first_last, every_other_rem, first_last_in_bet, reverse_slic, third]
-    test_lst = ['', 'tesu', 'ab',''.join(map(str,list(range(10)))), [1,2,3], list(range(24)), list(range(9)),['1third', '2third', '3third'], []]
+    functions = [exch_first_last, every_other_rem,
+                 first_last_in_bet, reverse_slic, third]
+    test_lst = ['', 'tesu', 'ab', ''.join(map(str, list(range(10)))),
```
PythonCHB, line 55: this is just a style issue, but I think it would be clearer if you put each on one line:

```
test_lst = ['',
                 'tesu',
                 'ab',
                 ''.join(map(str, list(range(10)))),[1,2,3],
                 list(range(24)),
                 list(range(9)),
                 ['1third', '2third', '3third'],
                 [],
                 ]
```
It took me a while to see which belonged with which.

PythonCHB (approved):
Really nice work!

You are going to be bored this week in class :-(

PythonCHB:
And by the way -- your solution to rot13 was excellent.

PR #52, opened 2017-10-15
merge in series02 files

Ginsusamurai:
series, fizzbuzz and grid exercises

### DIFF
--- a/students/morgan/session02/grid.py
+++ b/students/morgan/session02/grid.py
@@ -1,0 +1,63 @@
+
+
+# print('+ ', '- ' * 4, '+ ','- ' * 4, '+')
+# print('| ', '  ' * 4, '| ','  ' * 4, '|')
+# print('| ', '  ' * 4, '| ','  ' * 4, '|')
+# print('| ', '  ' * 4, '| ','  ' * 4, '|')
+# print('| ', '  ' * 4, '| ','  ' * 4, '|')
+# print('+ ', '- ' * 4, '+ ','- ' * 4, '+')
+# print('| ', '  ' * 4, '| ','  ' * 4, '|')
+# print('| ', '  ' * 4, '| ','  ' * 4, '|')
+# print('| ', '  ' * 4, '| ','  ' * 4, '|')
+# print('| ', '  ' * 4, '| ','  ' * 4, '|')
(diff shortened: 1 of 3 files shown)


students/morgan/session02/grid.py
```diff
@@ -0,0 +1,41 @@
+
+
+corner  = "+ "
```
PythonCHB, line 3: I like this!

Did you experiment with putting different characters in there to see how well it worked?

students/morgan/session02/series.py
```diff
@@ -0,0 +1,40 @@
...
+
+def lucas(n):
+	""" lucas number sequence. same as fib but starts with 2 and 1"""
+	if n == 0:
+		return 2
+	if n == 1:
+		return 1
+	return lucas(n-1) + lucas(n-2)
+
+def sum_series(n, b=0, c=1):
```
PythonCHB, line 17: Perfect!

students/morgan/session02/series.py
```diff
@@ -0,0 +1,40 @@
...
+	print(fib(2))
+
+	assert lucas(0) == 2
+	assert lucas(1) == 1
+	print(lucas(2))
+
+	assert sum_series(0,3,8) == 3
+	assert sum_series(1,3,8) == 8
+	print(sum_series(2,3,8))
+
```
PythonCHB, line 39: you want to make sure to test a couple larger values, not just the initial ones.

PythonCHB (approved):
This all looks good.

EXCEPT -- it looks like you are indenting with tabs. You really, really, don't want to do that.

The style standard for python code is to use 4 spaces for indenting -- always. If you are ever going to sahre code with anyone, you really, really want to use that standard.

Set up your editor to put in spaces when you hit the <tab> key, and you should be good to go.

students/morgan/session02/grid.py
Ginsusamurai: a bit. it looks fine, just not as even. +, -, | just look more square on the grid.

PythonCHB: yup -- those probably are the best options -- but goodto knowthat the does work with other characters..

PR #53, opened 2017-10-15
Revision on fibonacci code

HiroTestStudent:
I am trying to create pull request for my revised fibonacci code but is it working? It doesn't seem to be working...

Hiro

### DIFF
--- a/students/hirot_directory/fib_hiroTakechi.py
+++ b/students/hirot_directory/fib_hiroTakechi.py
@@ -4,8 +4,9 @@ Fibonacchi recursion Python code
 
 '''
 
-# in-class:
+# in-class (Chris's example):
 
+'''
 def fib(n):
 
 	if n == 0:
@@ -21,14 +22,17 @@ if __name__ == '__main__':
 	assert fib(3) == 2
 	print(fib(10))
 
+'''
+
 # the codes that I worked on after class: 
 
+'''
 ## code 1:
 def fibonacci(n, a=1, b=0):
     return b if n < 1 else fibonacci(n - 1, a + b, a)
 
 print(fibonacci(7))
-
+'''
 
 ## code 2:
 class Program():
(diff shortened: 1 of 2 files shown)


PythonCHB:
This PR seems to be working fine.

If you already have a PR open, then updates to your repo will automatically be added to the existing PR. So you may not have had to do anything.

students/hirot_directory/fib_hiroTakechi.py
```diff
@@ -21,14 +22,17 @@ def fib(n):
...
 ## code 1:
 def fibonacci(n, a=1, b=0):
     return b if n < 1 else fibonacci(n - 1, a + b, a)
 
 print(fibonacci(7))
-
+'''
 
 ## code 2:
 class Program():
```
PythonCHB, line 38: So this works, but is rally not how you'd want to do it.

In general, we don't put functions in classes by themselves at all in python.

We'll get to classes in a couple weeks.

students/hirot_directory/fizz.py
```diff
@@ -0,0 +1,51 @@
+
+'''
+Dev: Hiro Takechi
+Subject: FizzBuzz
+'''
+
+'''
+def print_fizz():
```
PythonCHB, line 8: breaking things down into functions is good in general, but probably not a good idea to do it when they are THIS simple.

PythonCHB: Though it looks like you didn't end up using them anyway.

PythonCHB (approved):
all good.

students/hirot_directory/fizz.py
HiroTestStudent: Hi Chris,

Yes, during Maria's office hours, she taught us to use def by showing the example of grid printer, so I just wanted to try to do the same approach to fizzBuzz but then I couldn't add 'n' in front of "fizz", "buzz" and "fizzBuzz" so I decided to comment them out. 

Thanks,
Hiro

students/hirot_directory/fib_hiroTakechi.py
HiroTestStudent: Hi Chris,

OK, then I will wait to learn about the class for a few weeks. About PR, I didn't see the same response from Github when I committed the revision so I wasn't sure of whether it was working but it did work. That's good news.

Thanks!
Hiro

PR #54, opened 2017-10-16
Adding Session02 Classwork 

AmitUW:
1. GridPrinter
2. FizzBuzz
3. Fibonacci

### DIFF
--- a/students/Amit/Session02/grid_printer.py
+++ b/students/Amit/Session02/grid_printer.py
@@ -9,6 +9,16 @@ minus = "-"
 pipe = "|"
 space = " "
 
+def grid():
+    line1 = plus+(3*(space+minus+space))
+    line2 = pipe+(3*(space+space+space))
+    for i in range(3):
+        print(line1*3 + plus)
+        for i in range(3):
+            print(line2*3 + pipe)
+    print(line1*3 + plus)
+
+
 def print_grid(n):
     line1 = plus+(n*(space+minus+space))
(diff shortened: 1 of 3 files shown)


students/Amit/Session02/grid_printer.py
```diff
@@ -0,0 +1,46 @@
+#! /usr/bin/env python
```
PythonCHB, line 1: this all looks good.

students/Amit/Session02/series.py
```diff
@@ -0,0 +1,22 @@
+#! /usr/bin/env python
+
+fib_cache= {}
+
+def fibonacci(n):
+    if n in fib_cache:
```
PythonCHB, line 6: very slick! This is called "memoization", and we'll get into a better way to do this in a few weeks.

but nice, and should really enhance performance, not computing the same one multiple times.

students/Amit/Session02/series.py
```diff
@@ -0,0 +1,22 @@
...
+        return fib_cache[n]
+
+    if n == 0:
+        return 0
+    elif n == 1:
+        return 1
+    elif n == 2:
```
PythonCHB, line 13: no need for this check -- the algorithm will work for n == 2 -- and you'll be caching the result anyway.

PythonCHB (approved):
nice work.

Maybe do the more generic sum_series next -- that will be trickier to cache -- but it can be done pretty simply, too.

PR #55, opened 2017-10-16
grid printer code

HiroTestStudent:
Hi Chris or Maria,

This is my grid printer code. I used hints from Maria's office hours on Saturday.

Thanks,
Hiro (Takechi)

### DIFF
--- a/students/hirot_directory/grid_printer.py
+++ b/students/hirot_directory/grid_printer.py
@@ -1,0 +1,61 @@
+
+'''
+Dev: Hiro Takechi
+Subject: Grid printer
+Date: 10-15-2017
+'''
+
+plus = '+'
+minus = '-'
+space = ' '
+bar = '|'
+
+def h_line(x):
+	x = int((x-1)/2)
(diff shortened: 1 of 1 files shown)


students/hirot_directory/grid_printer.py
```diff
@@ -0,0 +1,61 @@
...
+Date: 10-15-2017
+'''
+
+plus = '+'
+minus = '-'
+space = ' '
+bar = '|'
+
+def h_line(x):
+	x = int((x-1)/2)
```
PythonCHB, line 14: you can use the // operator, known as "floor division" it gives an integer result, so no need for the call to int().

PythonCHB (approved):
Nice work.

One suggestion -- if you have multiple funcitons to buildportions like this, it's better for them to return strings, and then have the calling code to the printing.

That way, it could be easily extended to write it to a file, or??? It's also easier to test.

No need to change this, but keep that in mind next week for the "Mailroom" project :-)

PR #56, opened 2017-10-16
John N's session 1 warmup scripts.

johnnzz:
I've been out of town on work and am trying to catch up.  Here are my session 1 warm-up scripts.

### DIFF
--- a/students/john_navitsky/session01/warmup1/break_me.py
+++ b/students/john_navitsky/session01/warmup1/break_me.py
@@ -1,0 +1,20 @@
+
+def NameError():
+    print(something_not_defined)
+
+def TypeError():
+    stringvalue="foo"
+    print(stringvalue + 10)
+
+def SyntaxError():
+    i_am_a_syntax_error
+
+def AttributeError():
+    integervalue=1
+    print(integervalue.upper())
+
+#NameError()
+#TypeError()
+#SyntaxError()
+AttributeError()
+
--- a/students/john_navitsky/session01/warmup1/front3.py
+++ b/students/john_navitsky/session01/warmup1/front3.py
@@ -1,0 +1,14 @@
+
+def front3(string):
+    index = 3
+    length = len(string)
+    if length <= 3:
+      index = length
+    front = string[0:index]
+    return front
+
+print(front3("Java"))
+print(front3("Chocolate"))
+print(front3("abc"))
+#print(front3("a"))
+#print(front3("ab"))
--- a/students/john_navitsky/session01/warmup1/front_back.py
+++ b/students/john_navitsky/session01/warmup1/front_back.py
@@ -1,0 +1,14 @@
+
+def front_back(string):
+    string_list=list(string)
+    index_last_char=len(string_list)-1
+    first_char=string_list[0]
+    last_char=string_list[index_last_char]
+    string_list[0]=last_char
+    string_list[index_last_char]=first_char
+    new_string="".join(string_list)    
+    return new_string
+
+print(front_back("code"))
+print(front_back("a"))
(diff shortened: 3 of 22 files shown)


students/john_navitsky/session01/warmup1/break_me.py
```diff
@@ -0,0 +1,20 @@
...
+
+def AttributeError():
+    integervalue=1
+    print(integervalue.upper())
+
+#NameError()
+#TypeError()
+#SyntaxError()
+AttributeError()
+
```
PythonCHB, line 20: All good!

students/john_navitsky/session01/warmup1/front3.py
```diff
@@ -0,0 +1,14 @@
+
+def front3(string):
+    index = 3
+    length = len(string)
+    if length <= 3:
+      index = length
+    front = string[0:index]
+    return front
+
```
PythonCHB, line 9: logic is fine, but python slicing can make this MUCH easier:

```
def front3(string):
    return 3 * string[:3]
```
(untested)

PythonCHB: NOTE: it's pretty rare to need to use the length for indexing when slicing....

students/john_navitsky/session01/warmup1/front_back.py
```diff
@@ -0,0 +1,14 @@
+
+def front_back(string):
```
PythonCHB, line 2: also can be much easier with slicing....

students/john_navitsky/session01/warmup1/makes10.py
```diff
@@ -0,0 +1,12 @@
+
+def makes10(num1, num2):
+    num_sum = num1 + num2
+    if num1 == 10 or num2 == 10 or num_sum == 10:
```
PythonCHB, line 4: this can be simplified by returning the result of the tests directly:

```
return num1 == 10 or num2 == 10 or num_sum == 10
```

students/john_navitsky/session01/warmup1/missing_char.py
```diff
@@ -0,0 +1,9 @@
+def missing_char(word, index):
+    if word:
+        wordlength=len(word)
+        start=0
+        return word[start:index]+word[index+1:wordlength]
```
PythonCHB, line 5: you dont need start or wordlength here -- they can be assumed:

```
return word[:index]+word[index+1:]
```

pretty nifty, eh?

students/john_navitsky/session01/warmup1/monkey_trouble.py
```diff
@@ -0,0 +1,13 @@
+
+def monkey_trouble(a_smile=False, b_smile=False):
+
+    bool_monkey=a_smile + b_smile
+    if bool_monkey in [2, 0]:
```
PythonCHB, line 5: slick! 

I wouldn't call the variable "bool_monkey", as it's not really a bool anymore. But I like the logic!

students/john_navitsky/session01/warmup2/string_times.py
```diff
@@ -0,0 +1,11 @@
+
+def string_times(string="",copies=1):
```
PythonCHB, line 2: you can multiply python strings, so this one is pretty trivial:

```
return string * copies
```
:-)

PythonCHB (approved):
nice job -- got the logic down.

Do think about about nifty things in python like slicing, etc. that can make these simpler and cleaner.

PythonCHB:
Do try to get to the reading, so you'll be ready to jump in tomorrow!

PR #57, opened 2017-10-16
adding some examples

alinafe82:


students/alinafe/codingBat/caught_speeding.py
```diff
@@ -0,0 +1,30 @@
...
+You are driving a little too fast, and a police officer stops you. Write code to compute the result, encoded as an int value: 0=no ticket, 1=small ticket, 2=big ticket. If speed is 60 or less, the result is 0. If speed is between 61 and 80 inclusive, the result is 1. If speed is 81 or more, the result is 2. Unless it is your birthday -- on that day, your speed can be 5 higher in all cases.
+"""
+
+def caught_speeding(speed, is_birthday):
+  if is_birthday:
+    speed -= 5
+     
+  if speed <= 60:
+      return 0
+  if 60 < speed <= 80:
```
PythonCHB, line 12: You've already checked that speed > 60 -- so this could be simpler.

not a big deal though -- sometimes repetition is better for clarity.

students/alinafe/codingBat/caught_speeding.py
```diff
@@ -0,0 +1,30 @@
...
+    return 1
+  return 2
+
+if __name__ == "__main__":
+    # some tests
+    caught_speeding(60, False)	is 0
```
PythonCHB, line 18: you want to use "assert" here -- this is OK, but you have to scan and make sure they all rturn True -- whereas an assert will raise an error for you if it fails.

PythonCHB, line 18: You don't want to use "is" here -- you want ==.

"is" tests if it is the same object. As it happens, cPython re-uses small integers, so one zero actually IS the same object as another zero -- but that's an accident of implementation.

I'm going to talk about that in class this week.

students/alinafe/codingBat/love6.py
```diff
@@ -0,0 +1,12 @@
...
+The number 6 is a truly great number. Given two int values, a and b, return True if either 
+one is 6. Or if their sum or difference is 6. Note: the function abs(num) computes the 
+absolute value of a number.
+"""
+def love6(a, b):
+  return a == 6 or b == 6 or (a + b) == 6 or abs(a - b) == 6
+What, this looks ugly you say? I completely agree, and there is a much more pleasant solution:
+
+def love6(a, b):
+  return 6 in [a, b, a + b, abs(a - b)]
```
PythonCHB, line 12: nice!

students/alinafe/codingBat/sorta_sum.py
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env python3
+def sorta_sum(a, b):
+  if 10 <= a + b < 20:
+    return 20
+  return a + b
+
+if __name__ == "__main__":
+    assert sorta_sum(3, 4)	is 7
+    assert sorta_sum(9, 4)	is 20
+    assert sorta_sum(10, 11) is	21
```
PythonCHB, line 10: what's all that whitespace? tab characters? avoid those!

students/alinafe/codingBat/squirrel_play.py
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env python3
+"""
+The squirrels in Palo Alto spend most of the day playing. In particular, they play if the temperature is between 60 and 90 (inclusive). Unless it is summer, then the upper limit is 100 instead of 90. Given an int temperature and a boolean is_summer, return True if the squirrels play and False otherwise.
+"""
+def squirrel_play(temp, is_summer):
+    return (temp >= 60 and temp <= 90) if not is_summer else (temp >= 60 and temp <= 100)
```
PythonCHB, line 6: nice!

students/alinafe/session02/fizz_buzz.py
```diff
@@ -0,0 +1,46 @@
...
+def fizzbuzz(n):
+    for n in range(1,n+1):
+        output=""
+        if n % 3 == 0:
+            output += "Fizz"
+        if n % 5 == 0:
+            output += "Buzz"
+        if not output:
+            output = n
+            print(output)
```
PythonCHB, line 42: you don't want the print line indented here -- it should always print.

students/alinafe/session02/print_grid.py
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env python
+"""
+Grid Printer Exercise:
+"""
+
+def print_grid():
+
```
PythonCHB, line 7: this is fine, but does violate the DRY principle (don't repeat yourself).
soring the horizontal and vertical lines in a variable would be a way to avoid that.

students/alinafe/session02/series.py
```diff
@@ -0,0 +1,25 @@
+#series.py
+
+#def F():
```
PythonCHB, line 3: whoa! a generator!

We'll get to that later in the class :-)

students/alinafe/session02/series.py
```diff
@@ -0,0 +1,25 @@
...
+        return 0
+    elif n == 1:
+        return 1
+    else:
+        return fibonacci(n-1)+fibonacci(n-2)
+
+if __name__=='__main__':
+    assert fibonacci(0) ==0
+    assert fibonacci(1) ==1
+    assert fibonacci(100) ==1
```
PythonCHB, line 20: huh? fibonacci(100) should be a huge number....

PythonCHB (commented):
nice work.

I like the various examples -- really good to try stuff out to make sure you know how it works.

PR #60, opened 2017-10-17
John N's session02 assignment

johnnzz:
fizz_buzz.py
  Identify multiples of 3 and 5

print_grid.py
  Print user defined squares

series.py
  Return Nth numbers from a sum sequence

### DIFF
--- a/students/john_navitsky/session02/fizz_buzz.py
+++ b/students/john_navitsky/session02/fizz_buzz.py
@@ -1,0 +1,21 @@
+
+""" Print fizz or buzz if the value is a multiple of 3 or five, respectively. """
+
+for i in range(1, 101):
+
+    message=""
+
+    # is the number a multiple of three?
+    if int( i / 3 ) == float( i / 3 ):
+        message=message+"Fizz"
+
+    # is the number a multiple of five?
+    if int( i / 5 ) == float( i / 5 ):
+        message=message+"Buzz"
+     
+    if not message:
+        print(i)
+    else:
+        print(message)
+        #print(message,"(",i,")")
+
--- a/students/john_navitsky/session02/print_grid.py
+++ b/students/john_navitsky/session02/print_grid.py
@@ -1,0 +1,34 @@
+
+""" A collection of functions to print user definable text squares. """
+
+def print_grid(x_squares=1,y_squares=1,square_size=1):
+    """ Print a grid given square size, number of x and y squares. """
+    #print("x_squares:",x_squares,"y_squares:",y_squares,"square_size:",square_size)
+    # number of y squares
+    for y_row in range(y_squares):
+        # print the top
+        for top in range(x_squares):
+            print("+" + "-" * square_size, end= "" )
+        print("+")
+        # print the middle lines
+        for middle in range(square_size):
+            for top in range(x_squares):
+                print("|" + " " * square_size, end= "" )
+            print("|")
+    # print the bottom
+    for bottom in range(x_squares):
+        print("+" + "-" * square_size, end= "" )
+    print("+")
+
+def print_grid_fixed(square_size=3):
+    """ Print a grid given square size. """
+    print_grid(2,2,square_size)
+
+def print_grid_cells(num_cells=2,square_size=2):
+    """ Print a grid number of squares and square size. """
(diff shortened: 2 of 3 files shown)


students/john_navitsky/session02/fizz_buzz.py
```diff
@@ -0,0 +1,21 @@
+
+""" Print fizz or buzz if the value is a multiple of 3 or five, respectively. """
+
+for i in range(1, 101):
+
+    message=""
+
+    # is the number a multiple of three?
+    if int( i / 3 ) == float( i / 3 ):
```
PythonCHB, line 9: you don't need the "float" the results of division are always a float.

and you don't need the "int" either, as you can use the "floor division" operator (double slash): `i // 3`

but you do'nt need any of that, as there is the "mod" or remainder operator: `%`

so this becomes:

```
if i % 3 == 0:
```

or even:

```
if not i % 3:
```
(as anything other than zero is True)

students/john_navitsky/session02/fizz_buzz.py
```diff
@@ -0,0 +1,21 @@
...
+        message=message+"Fizz"
```
PythonCHB, line 10: this can be: `message += "Fizz"

students/john_navitsky/session02/fizz_buzz.py
```diff
@@ -0,0 +1,21 @@
...
+    # is the number a multiple of five?
+    if int( i / 5 ) == float( i / 5 ):
+        message=message+"Buzz"
+     
+    if not message:
+        print(i)
+    else:
+        print(message)
+        #print(message,"(",i,")")
+
```
PythonCHB, line 21: log is is good !

students/john_navitsky/session02/print_grid.py
```diff
@@ -0,0 +1,34 @@
+
```
PythonCHB, line 1: This all looks good.

students/john_navitsky/session02/series.py
```diff
@@ -0,0 +1,49 @@
+
+""" Functions that return Nth entry of a sum sequence. """
+
+def sum_series(nth=1,sequence=[0,1]):
```
PythonCHB, line 4: This is the non-recursive way to do it -- which, in fact, is probably better performing (though if you don't need the whole sequence, it may store extraneous stuff.

But do make sure you know how to do it recursively as well -- sometimes that makes the logic much simpler.

PythonCHB, line 4: Careful here!

This will work if you call it once, but can give very wrong results called additional times. Try:

assert sum_series(5) == 3
assert sum_series(6) == 5

It fails! To see what's going on, print the sequence:

```
In [17]: def sum_series(nth=1,sequence=[0,1]):
    ...:     """ Generate a list of sums given a seed and return the Nth number.
    ...:  """
    ...:     for i in range(2, nth):
    ...:         sequence.append(sequence[i-2]+sequence[i-1])
    ...:     print(sequence)
    ...:     return sequence[nth-1]
    ...: 

In [18]: sum_series(5)
[0, 1, 1, 2, 3]
Out[18]: 3

In [19]: sum_series(6)
[0, 1, 1, 2, 3, 1, 2, 3, 5]
Out[19]: 1
```
Why???

We'll get into this in class later, but that default argument: `sequence=[0,1]` gets evaluated ONCE, when the function is defined. so every time the function is used, it re-uses the same list, and keeps adding to it!

As a rule, you don't want to use a mutable object as a default argument. Rather, something like this:

```
def sum_series(nth=1, sequence=None):
    if sequence is None:
        sequence=[0,1]
```

that will guarantee you get a new one each time the function is run.

Another, maybe better option is to use an immutable sequence:

```
def sum_series(nth=1, sequence=(0, 1)):
    sequence = list(sequence)
```
That way, the user can pass in any sequence (tuple or list), and if they pass in a list it won't be mutated by the function.

PythonCHB (approved):
Nice work -- that mutable default argument is a real "gotcha" in Python! I wouldn't have expected anyone new to Python to see it.

PythonCHB:
You are welcome to update and submit a new PR if you want further review, but not necessary.

PR #65, opened 2017-10-20
Session 3 exercises

Eowyn42:
Completed all session 3 exercises including the "optional" encoding exercise (which is extremely simple after exploring the docs). I'm continuing to look for ways to improve the mailroom code by abstracting tasks into sub functions. Happy to receive feedback now on this code. Thank you!

### DIFF
--- a/students/eowyn/session03/list_lab.py
+++ b/students/eowyn/session03/list_lab.py
@@ -38,8 +38,7 @@ def remove_disliked_fruit():
     i = 0
     stop = len(fruits)
     while i < len(fruits):
-        fruit = fruits[i]
-        like = input("Do you like {}, yes or no? ".format(fruit.lower()))
+        like = input("Do you like {}, yes or no? ".format(fruits[i].lower()))
         if like.lower() == "no":
             del fruits[i]
         elif like.lower() == "yes":
@@ -51,15 +50,19 @@ def remove_disliked_fruit():
 
 
 
-
-
-
 def rev_fruitnames():
-    print("copy list and reverse letters of each fruit")
-    print("then delete the last item of original list")
+    '''copy list and reverse letters of each fruit
+    then delete the last item of original list'''
+    fruits = ['Apples','Pears','Oranges','Peaches']
+    fruits_copy = fruits[:]
+    fruits_copy = [x[::-1] for x in fruits_copy]
+    del fruits[-1]
+    print(fruits)
+    print(fruits_copy)
+
 
 if __name__=="__main__":
   #  list_fruit()
   #  remove_fruit()
-    remove_disliked_fruit()
+  #  remove_disliked_fruit()
     rev_fruitnames()
--- a/students/eowyn/session03/list_lab.py
+++ b/students/eowyn/session03/list_lab.py
@@ -62,7 +62,11 @@ def rev_fruitnames():
 
 
 if __name__=="__main__":
-  #  list_fruit()
-  #  remove_fruit()
-  #  remove_disliked_fruit()
+    print("running list_fruit")
+    list_fruit()
+    print("running remove_fruit")
+    remove_fruit()
+    print("running remove_disliked_fruit")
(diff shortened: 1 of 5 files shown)


students/eowyn/session03/list_lab.py
```diff
@@ -0,0 +1,72 @@
...
+    fruits = ['Apples','Pears','Oranges','Peaches']
+    print(fruits)
+    del fruits[-1]
+    print(fruits)
+    choice = input("Which fruit should be removed? > ")
+    print(choice,type(choice))
+    ## this works in ipython but not here...?
+    ## should remove all instances of choice
+    fruits = [x for x in fruits if x!= choice]
+
```
codedragon, line 33: Why do you think it isn't working here? Try printing fruits in the function after you remove one.

students/eowyn/session03/mailroom.py
```diff
@@ -0,0 +1,107 @@
...
+        fullname = remove_inputquotes(fullname)
+        if fullname == 'menu':
+            break
+        if fullname.lower() == 'all':
+            print("All Donors:")
+            [print(x[0]) for x in DONORS]
+        else:
+            try: 
+                i = [x[0] for x in DONORS].index(fullname)
+                print("Existing Donor")
```
codedragon, line 43: Nice list comprehension. :) It is useful to only have one thing in a try block, that way you know precisely where the failure is. Note that you are writing a couple of lines of code twice, so no reason to have those in the try/except block anyway. You can use the else block to keep going. Example:
```
amount = float(input("Donation amount: "))
try:
    i = [x[0] for x in DONORS].index(fullname)
except ValueError:
   DONORS.append((fullname,[amount]))
   this_donor = DONORS[-1]
else:
    this_donor = DONORS[i]
    this_donor[1].append(amount)
print_letter(this_donor)    
```

students/eowyn/session03/mailroom.py
```diff
@@ -0,0 +1,107 @@
...
+    ''' Generate a formatted thank you note to a donor '''
+    fs = "Thank you, {}, for your generosity and recent gift of ${}."
+    print(fs.format(a_donor[0],a_donor[1][-1]))
+
+def remove_inputquotes(a_string):
+    '''check for and remove anxillary quotes'''
+    if a_string[0]==a_string[-1]=='"' or a_string[0]==a_string[-1]=="'":
+        a_string = a_string[1:-1]
+    return a_string
+
```
codedragon, line 91: This remove_inputquotes function should not be necessary. What was happening that made you think you needed it?

students/eowyn/session03/rot13.py
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env python
+
+def rot13(a_string):
+    ''' Encode any string using rot13 encoding, preserving 
+    whitespace, punctuation, and capitalization '''
+    import codecs
+    return codecs.encode(a_string,encoding="rot13")
+
```
codedragon, line 8: Nice catch. There are other ways to do this, and not everyone notices there is a rot13 codec. :)

students/eowyn/session03/mailroom.py
```diff
@@ -0,0 +1,107 @@
...
+            for x in range(len(DONORS)):
+                this_donor = DONORS[x]
+                name.append(this_donor[0])
+                total.append(round(sum(this_donor[1])))
+                number.append(len(this_donor[1]))
+                ave.append(round(total[x]/number[x]))
+            report_data = list(zip(name,total,number,ave))
+            report_data = sorted(report_data,key = lambda y:int(y[1]),reverse = True)
+            print_table(report_data)
+
```
codedragon, line 77: You could have the code for creating the actual report_data in a separate function, so the else statement may just be:
```
else:
    report_data = get_report_data()
    print_table(report_data)
```
Nice use of sorted.

codedragon (approved):
Great job!

PR #67, opened 2017-10-22
Session03

Ginsusamurai:
mailroom exercise completed.

### DIFF
--- a/students/morgan/session03/list_lab.py
+++ b/students/morgan/session03/list_lab.py
@@ -1,0 +1,87 @@
+#!/usr/bin/env python3
+# $ ./list_lab.py
+# chmod +x list_lab.py
+
+fruit = ['Apples', 'Pears', 'Oranges', 'Peaches']
+
+print(fruit)
+
+new_fruit = input("What's fruit would you like to add?\n"
+            "> ")
+
+fruit.append(new_fruit)
+
+print(fruit)
+    
+number = int(input("What's your number?\n"
+            "> "))
+
+print(number, fruit[number-1])
+
+newest_fruit = []
+newest_fruit.append(input("What's fruit would you like to add?\n"
+            "> "))
+print(newest_fruit)
+added_fruit_list = newest_fruit + fruit
+print(added_fruit_list)
+
+added_fruit_list.insert(0, input("What's fruit would you like to add?\n"
+            "> "))
+
+for x in range(len(added_fruit_list)):
+    if str.lower(added_fruit_list[x][0:1]) == 'p':
+        print(added_fruit_list[x])
+
+
+# Series 2
+
+print(added_fruit_list)
+
+added_fruit_list = added_fruit_list[:-1]
+print(added_fruit_list)
+
+
+remove_fruit = input("What's fruit would you like to remove?\n"
+            "> ")
+
+print(remove_fruit)
+print(added_fruit_list[0])
+
+for thing in added_fruit_list:
+    if str.lower(thing) == str.lower(remove_fruit):
+        added_fruit_list.remove(thing)
+
+print(added_fruit_list)
+
+# Series 3
+
+on = True
+for thing in added_fruit_list[:]:
+    
+    while on == True:
+        x = input('Do you like ' + thing + '?\n')
+        if str.lower(x) == 'no':
+            added_fruit_list.remove(thing)
+            break
+        elif str.lower(x) == 'yes':
+            break
+        else:
+            print("Please enter 'yes' or 'no'\n")
+
+print(added_fruit_list)
+
+# Series 4
+
+
+reverse_fruit = fruit[:]
+
+i = 0
+while i < len(reverse_fruit):
+    reverse_fruit[i] = reverse_fruit[i][::-1]
+    i += 1
+
+del fruit[-1:]
+print(fruit)
+print(reverse_fruit)
+
+input('wait')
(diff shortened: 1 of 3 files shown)


students/morgan/session03/list_lab.py
```diff
@@ -0,0 +1,87 @@
...
+added_fruit_list = newest_fruit + fruit
+print(added_fruit_list)
+
+added_fruit_list.insert(0, input("What's fruit would you like to add?\n"
+            "> "))
+
+for x in range(len(added_fruit_list)):
+    if str.lower(added_fruit_list[x][0:1]) == 'p':
+        print(added_fruit_list[x])
+
```
codedragon, line 34: You don't need the index, believe it or not:
```
for fruit in added_fruit_list:
    if str.lower(fruit[0]) == 'p':
        print(fruit)
```

students/morgan/session03/list_lab.py
```diff
@@ -0,0 +1,87 @@
...
+remove_fruit = input("What's fruit would you like to remove?\n"
+            "> ")
+
+print(remove_fruit)
+print(added_fruit_list[0])
+
+for thing in added_fruit_list:
+    if str.lower(thing) == str.lower(remove_fruit):
+        added_fruit_list.remove(thing)
+
```
codedragon, line 53: Be very careful deleting objects from a list that you are iterating through. Mostly, don't do it. An interesting alternative Chris came up with:
```
while ans in d_fruits:
    d_fruits.remove(ans)
```
This doesn't address the lowercase issue, however. Creating a copy of the list, minus the fruit to be removed is one option:
```
new_list = [fruit for fruit in added_fruit_list if str.lower(fruit) != str.lower(remove_fruit)]
```

students/morgan/session03/list_lab.py
```diff
@@ -0,0 +1,87 @@
...
+    while on == True:
+        x = input('Do you like ' + thing + '?\n')
+        if str.lower(x) == 'no':
+            added_fruit_list.remove(thing)
+            break
+        elif str.lower(x) == 'yes':
+            break
+        else:
+            print("Please enter 'yes' or 'no'\n")
+
```
codedragon, line 70: Since you don't actually use the variable on anywhere, you could just say `while True:` In fact, you don't need the while loop at all, you could put `continue` or `pass` for the yes option, and it would just skip on to the next fruit.

students/morgan/session03/list_lab.py
```diff
@@ -0,0 +1,87 @@
...
+# Series 4
+
+
+reverse_fruit = fruit[:]
+
+i = 0
+while i < len(reverse_fruit):
+    reverse_fruit[i] = reverse_fruit[i][::-1]
+    i += 1
+
```
codedragon, line 82: nice solution

students/morgan/session03/mailroom.py
```diff
@@ -0,0 +1,129 @@
...
+              "> "))
+        if answer =='3':
+            break
+        elif answer =='1':
+            thank_you()
+        elif answer =='2':
+            print_report(donor_list)
+        else:
+            print("\nPlease type 1, 2, or 3")
+
```
codedragon, line 21: Note that this function is only going to work if this file is run as a script. If you try to import these to another file or the command line, this function will have no idea what donor_list is, since you only define it in the `if __name__ == "__main__"` block.

students/morgan/session03/mailroom.py
```diff
@@ -0,0 +1,129 @@
...
+            print("\nPlease type 1, 2, or 3")
+            
+
+    
+
+def print_report(donors):
+    '''need the list of donors sorted before I can print them accurately,
+    found the sorted function while looking for a way to do this'''
+    donors = sorted(donors, key=lambda x: x[1], reverse=True)
+
```
codedragon, line 29: Nice find. :)

students/morgan/session03/mailroom.py
```diff
@@ -0,0 +1,129 @@
...
+    elif name_in_list(donor_input):
+        '''if the donor exists in the list, will fetch the index of that entry
+        and add the donation to the donors list and increment the number of donations.
+        I wanted to make sure the correctly capitalized name was kept and used
+        donation converted to float for safety'''
+        proper_name = donor_list[name_in_list(donor_input)][0]
+        donor_donation = float(input('Enter donation amout \n> '))
+        add_donation(proper_name, donor_donation)
+        write_letter(proper_name, donor_donation)
+
```
codedragon, line 69: So, usually the reason that you would ask to see all of the donors, is so that you are sure you are writing the name correctly, so you would expect that after asking for the names, you would be prompted again for the thank you note. So, you might want to consider putting in a loop and including exiting to main menu as one of the choices. Once you do this, you may find you can pull stuff in the if blocks into separate functions. 
Notice also that you have typed the line `donor_donation = float(input... ` and the line write_letter twice. DRY (don't repeat yourself) is a principle we will be talking more about...

students/morgan/session03/mailroom.py
```diff
@@ -0,0 +1,129 @@
...
+    ['dude',200.00, 2],
+    # 100
+    ['Paul Allen', 708.42, 3]
+    # 236.14
+    ]
+     
+    print('starting...')
+    main_loop()
+
+
```
codedragon, line 127: Nice job. You have mostly nice short functions, and hopefully I've given you some direction to fix the thank_you function.

codedragon (approved):
Looks great. :)

students/morgan/session03/list_lab.py
Ginsusamurai: i'll play with continue more but it was sticking in a loop when I tried that.

PR #68, opened 2017-10-22
Session03 Codes

chops77:


students/Chao/session03/list_lab.py
```diff
@@ -0,0 +1,99 @@
...
+def series_1():
+    """ Series 1 tasks """
+    
+    print("The current fruit list: " + str(fruits))
+    
+    # Append a fruit
+    morefruit = input("Enter the fruit you want to add to the list: ")
+    fruits.append(morefruit)
+    print("The current fruit list: " + str(fruits) + '\n')
+
```
codedragon, line 15: Cool, did not realize you could do this. Alternatively, if you want the list of fruits to be more sentence-like you could do:
```
print("The current fruit list: " + ", ".join(fruits) + '\n')
```

students/Chao/session03/list_lab.py
```diff
@@ -0,0 +1,99 @@
...
+    print("The current fruit list: " + str(fr2))
+
+    while True:
+        removefruit = input("Please enter the fruit name you would like to remove: ")
+        if removefruit not in fr2:
+            print("No match, please try again.")
+        else:
+            while removefruit in fr2:
+                fr2.remove(removefruit)
+            break
```
codedragon, line 54: Good way to remove from a list. Makes sure you get all of them, and not iterating through the list while removing.

students/Chao/session03/list_lab.py
```diff
@@ -0,0 +1,99 @@
...
+        while True:
+            result = input("Do you like " + fruit + "? ")
+            if result == "yes":
+                break
+            elif result == "no":
+                fruits.remove(fruit)
+                break 
+            else:
+                print("Please type in yes or no only.")
+
```
codedragon, line 72: It is not necessary to have the while loop in the for loop. The for loop is sufficient, just use continue or pass in the result == 'yes' block to skip to the next item in the list.

students/Chao/session03/list_lab.py
```diff
@@ -0,0 +1,99 @@
...
+    for l in range(len(newfruits)):
+        newfruits[l] = newfruits[l][::-1]
+
+    # Remove the last item from original list
+    del fruits[-1]
+
+    # Compare the result
+    print("Original list: " + str(fruits))
+    print("Copied list: " + str(newfruits))
+
```
codedragon, line 92: Nicely done in this and the previous series. You seem to have the idea of how and why to make copies of lists.

students/Chao/session03/mailroom.py
```diff
@@ -0,0 +1,95 @@
...
+        else:
+            print("\n*** Selected item not in the menu. Please try again. ***\n")
+
+def donor_search(n):
+    """ Search donor by name, return donor tuple """
+    for donor in donors:
+        if donor[0].lower() == n.lower():
+            return donor
+    return None
+
```
codedragon, line 30: The return None is not strictly necessary, since that is the default.

students/Chao/session03/mailroom.py
```diff
@@ -0,0 +1,95 @@
...
+            print("Donor record added - Name: {}, Donation: ${}".format(donors[-1][0], donors[-1][1][-1]))
+            print("\nSending Thank You email...")
+            email(result)
+        # Existing donor, add new donation
+        else:
+            amount = input("Existing donor: {}\nPlease enter the amount of donation: ".format(donor_search(result)))
+            donor_search(result)[1].append(float(amount))
+            print("New Donation Added: {}".format(donor_search(result)))
+            print("\nSending Thank You email...")
+            email(result)
```
codedragon, line 70: Notice how much of the above two blocks of code are the same. How could you avoid repeating yourself?

students/Chao/session03/mailroom.py
```diff
@@ -0,0 +1,95 @@
...
+
+        # Print report
+        print("* {: <30}{:16.2f}{: >16}{:18.2f} *".format(donor[0], total, don, avg))
+    print("====================================================================================")
+
+
+
+if __name__ == '__main__':
+    """ Main function """
+    main_loop()
```
codedragon, line 95: Nice solution to the mailroom exercise!

codedragon (approved):
Very nice work!

students/Chao/session03/list_lab.py
codedragon: Or you can also do `print("The current fruit list: ", fruits, "\n")`

PR #69, opened 2017-10-23
Updates to Session 02 and Session 03 Files

brhelodrvr:
New edit for 10/23:
- If I've done this correctly, series.py from Session 02 should be updated to the newest version. 
- In series.py, I'm having trouble getting the sum_series function to format the Lucas Numbers result (Line 47 in code) correctly in the output on Line 58 (i.e. 'print("Output is:", output) in run_program function. 

Previous comments:
1. I need some help with the arbitrary-number section Session03 string_formatting exercise.  I found *a* way to do it that is not *the* way intended by this assignment.  I'd appreciate feedback regarding the better/correct way to approach it.
2. I couldn't figure out how to approach the mailroom exercise as a list-only answer.  I resorted to dictionaries to complete it.
3. Some of the Session 02 files are still in-work.  I'm catching up in the wake of my family-emergency absences over the last couple of weeks.  I hope to have these sorted out this week.


PythonCHB:
Brian,

We got it -- lots to review, so it may be a while :-)

-Chris

students/brian_warn/session02/series.py
```diff
@@ -0,0 +1,54 @@
...
+    def sum_series(n, a = 0, b = 1):
+        ''' Compute Lucas Numbers series or Fibonacci series based upon optional parameter values
+        :arg 
+            n = input integer
+            a = first optional argument
+            b = second optional argument
+        :return: nth value in either the Lucas Numbers series or Fibonacci series
+        '''
+        if a == 0 and b == 1:
+            result = 'Fibonacci Sequence Result for ' + str(n) + ' is :' + fibonacci(n)
```
PythonCHB, line 36: without running the code, I think I see the error.

you are trying to add together a bunch of strings, which wokds fine. But that last one is a problem:

```
  + fibonacci(n)
```
that function call returns an integer, not a string, so you can't add it to a string.

try wrapping it in a str() call, like you did for the "n" earlier.

Or better yet, replace that line with string formatting :-)

students/brian_warn/session02/fizzbuzz.py
```diff
@@ -0,0 +1,20 @@
...
+		# multiples of 3 --> fizz; multiples of 5 --> buzz; multiples of 3 and 5 --> fizz buzz
+		if i%3 == 0 and i%5 == 0:
+			print("fizzbuzz")
+		elif i%3 == 0 and i%5 != 0:
+			print("fizz")
+		elif i%3 != 0 and i%5 == 0:
+			print("buzz")
+		else:
+			print(i)
+
```
codedragon, line 18: Since you are checking for both 3 and 5 factors in the first if, it isn't necessary to check the not in the next two elif statements.

students/brian_warn/session03/mailroom.py
```diff
@@ -0,0 +1,112 @@
...
+            ty_name = str(input("Please enter the donor's name or type 'list' to see donor names. (Enter 'menu' to return to the original menu.) --> ")).strip()
+            if ty_name == 'list':
+                print("List of Donors:")
+                for ty_name in donor_gifts:
+                    print(ty_name)
+            elif ty_name == 'menu':
+                return
+            else:
+                break
+
```
codedragon, line 41: Good use of return and break

students/brian_warn/session03/mailroom.py
```diff
@@ -0,0 +1,112 @@
...
+            elif strUserChoice == '3':  # exit
+                print("Choice 3: Quitting")
+                break
+        return strUserChoice
+
+
+    show_menu()
+
+if __name__ == "__main__":
+    main()
```
codedragon, line 112: Nice mailman solution. As you discovered, dictionaries are the best data structure for this project. One way to do it as a list is to use two lists (zipped together if you want). One has the donor names, and the other is a list of lists, where each inner list corresponds to the donor, and has all donations for that donor in it. This is delicate, as it is easy for these lists to become out of sync, and then chaos! So, don't worry about figuring out the better data structure early. :)

students/brian_warn/session03/string_formatting.py
```diff
@@ -0,0 +1,30 @@
...
+print(new_tuple)
+
+# Take an arbitrary number of values
+
+vals = ast.literal_eval(input("Enter the comma-separated values for this example: "))
+
+print("vals is: ", vals)
+placeholder = '{:} '*len(vals)
+
+print("The",len(vals),"numbers are:",placeholder.format(*vals))
```
codedragon, line 30: Wow, I have never used ast, interesting find. So, we were expecting this to be a function, that would except a list or tuple of numbers. You figured out the difficult part that we envisioned. :) To change the string of numbers to a list of numbers, you can do this:
```
list_vals = []
for i in vals.split(','):
     list_vals.append(int(i))
```
Which, once we learn about list comprehensions, you will be able to do in one line. Now, you will be able to use your code to print out the expected string (substituting list_vals for vals, of course).

codedragon (approved):
Nice job!

PR #71, opened 2017-10-23
send_thank_you function

zandra2:
session03 (mailroom.py)

How could I have changed my send_thank_you function to do the following:
collect donor name, amount and append that information to generate a report with current and previous donors. I thought I have it, but apparently, it didn't do as I thought it would. After 2 days of trying I'm finally giving up.  

Thank you,

Zandra


zandra2:
Please review mailroom1.py instead of mailroom.py. I made some changes to mailroom.py and mailroom1.py is its replacement. Thank  you!

session03/mailroom1.py
```diff
@@ -0,0 +1,96 @@
...
+
+def find_donor(name):
+    """ finding donor name in donor_list and printing their information """
+    for donor in donor_list:
+        # case-insenstive compare
+        if name.lower() == donor[0].lower():
+            return donor
+    return None
+
+def sort_key(item):
```
PythonCHB, line 35: nicely done

session03/mailroom1.py
```diff
@@ -0,0 +1,96 @@
...
+    """ key function used to sort the list by first (not zeroth) item"""
+    return item[1]
+
+def print_letter(name, amount):
```
PythonCHB, line 39: looks good, though better for a function like this to return a string, and print it elsewhere. YOu'll see why when you add writing to a file this week :-)

session03/mailroom1.py
```diff
@@ -0,0 +1,96 @@
...
+    """ writing a thank you letter to donor """
+    print("\nDear {},\n\nThank you for your kind donation of ${}."
+          "\n\nSincerely,\n\nA Team".format(name, amount))
+
+def print_report():
+    """ printing a report that includes donor names, contribution amounts, sum, and average """
+    report = []
+    for (name, amount) in donor_list:
```
PythonCHB, line 47: think a bit about naming -- "amount" is singular, but in tis case it is a list of multiple numbers -- maybe "amounts"? or "donations"?

session03/mailroom1.py
```diff
@@ -0,0 +1,96 @@
...
+        total_amount = sum(amount)
+        number_gifts = len(amount)
+        avg_gift = total_amount / number_gifts
+        report.append((name, total_amount, number_gifts, avg_gift))
```
PythonCHB, line 51: very good!

session03/mailroom1.py
PythonCHB, line 35: maybe give this a name that lets you know about _how_ it is sorting? IN this case, it's the only sorting function, but in a larger program you may have more than one.

session03/mailroom1.py
```diff
@@ -0,0 +1,96 @@
...
+
+    # sorting report data A-Z
+    report.sort(key=sort_key)
+    # printing report in a nicer format
+    print("{:25s} | {:11s} | {:9s} | {:12s}".format(
+        "\nDonor Name", "Total amount", "# of Gifts", "Average amount"))
+    print("-" * 70)
+    for row in report:
+        print("{:25s}   {:11.2f}   {:9d}   {:12.2f}".format(*row))
```
PythonCHB, line 60: perfect!

session03/mailroom1.py
```diff
@@ -0,0 +1,96 @@
...
+            donor_list.append(donor)
+        else:
+            donor[1].append(float(amount))
+
+    # print a letter to thanks the donor
+    print(print_letter(answer, amount))
+
+if __name__ == "__main__":
+    run = True
+    while run:
```
PythonCHB, line 83: I prefer:

while True

and then a break in the "quit" stanza. But not that different.

PythonCHB (approved):
This all looks great -- nice work!

PR #76, opened 2017-10-24
adding rot13

alinafe82:
Just need the rot13 reviewed the rest is work in progress


PythonCHB:
whew! 'cause there is a lot here!

students/alinafe/session03/rot13.py
```diff
@@ -0,0 +1,85 @@
...
+
+"""
+
+#ord('A') = 65
+#ord('Z') = 90
+#ord('a') = 97
+#ord('z') = 122
+
+
+def setn(x, low, high):
```
PythonCHB, line 43: interesting -- I don't think I've seen it done quite this way before, but it sure works.

students/alinafe/session03/rot13.py
```diff
@@ -0,0 +1,85 @@
...
+    while x > high:
```
PythonCHB, line 44: a while here? are you going to have do it more than once?

I think somethign like:

```
x = x - 26 if x > high else x
```
would work too.

PythonCHB (approved):
looks good -- you got the idea of working with ordinals, and the logic is sound.

PR #78, opened 2017-10-25
Adding mailroom script for review

dkkan:


students/davidkan/session03/mailroom.py
```diff
@@ -0,0 +1,176 @@
...
+# You can store that data structure in the global namespace.
+# (2) The script should prompt the user (you) to chooose from a menu of 3 actions:
+# Send a Thank You", "Create a Report" or "Quit".
+# Comments: Remember to use verbs for function names and nouns for variables.
+# In addition, you can use a dictionary; however try to use list and tuples for this assignment
+# Last modified on 10/22/2017 by David Kan
+
+
+# Define global varables and data structure
+donors = []
```
PythonCHB, line 14: line 17 -- the list(zip()) call, will create a new list -- no need to create an empty one here first -- it willjsut get thrown away.

students/davidkan/session03/mailroom.py
```diff
@@ -0,0 +1,176 @@
...
+
+def show_donor_list():
+    """
+    Function will show all of the names in the donor list data structure
+    :return: None
+    """
+    print()
+    # List all names in the donor list data structure
+    for index in range(len(donors)):
+        print(''.join([str(index + 1),'. ', '{}'.format(donors[index][0])]))
```
PythonCHB, line 28: no need to combine join() and string formatting -- might as well do it all with format():

`print('{}. {}'.format(index +1, donor[0]))`

students/davidkan/session03/mailroom.py
```diff
@@ -0,0 +1,176 @@
...
+
+def add_donor(full_name):
+    """
+    Function will add a username to the donor list data structure if it does not already exist
+    :param full_name: username in lowercase
+    :return: None
+    """
+    print()
+    # If username not in list, then add that name to the data structure and use it
+    donors.append(('{}'.format(full_name), []))
```
PythonCHB, line 39: `'{}'.format(full_name)` this is doing the same thing as str(full_name), and full_name is already a  string, yes? 

So just:

`donors.append((full_name, []))`

would do fine

students/davidkan/session03/mailroom.py
```diff
@@ -0,0 +1,176 @@
...
+
+
+def get_donors_index(full_name):
+    """
+    Function will get the index location of the donor in the donor list data structure
+    :param full_name: username in lowercase
+    :return: donor_name_index - the index location of the donor
+    """
+    # If user types a name in the list, then use it
+    donor_name_index = [row[0] for row in donors].index(full_name)
```
PythonCHB, line 49: nice comprehension!
though it might make more sense to jsut loop through the donors looking for the one you want:

```
for donor in donors:
    if donor[0] == full_name:
        break
```
donor will now be the tuple of that donor.

students/davidkan/session03/mailroom.py
```diff
@@ -0,0 +1,176 @@
...
+    Function will add a username to the donor list data structure if it does not already exist
+    :param full_name: username in lowercase
+    :return: None
+    """
+    print()
+    # If username not in list, then add that name to the data structure and use it
+    donors.append(('{}'.format(full_name), []))
+
+
+def get_donors_index(full_name):
```
PythonCHB, line 42: no need really, to get the index of the donor -- you can jsut get the actual donor record (tuple in this case), and pass that back to use it.

You'd be amazed at how little you actually need indexes in Python!

students/davidkan/session03/mailroom.py
```diff
@@ -0,0 +1,176 @@
...
+    :return: donation_amount - the amount donated by the donor
+    """
+    print()
+    try:
+        # Once a name has been selected, prompt for a donation amount.
+        # Turn the amount into a number – it is OK at this point for the
+        # program to crash if someone types a bogus amount
+        donation_amount = int(input("Please specify the donation amount:\n"
+                                    "\n> "))
+    except (ValueError, UnboundLocalError) as e:
```
PythonCHB, line 66: nice!

Though when would you get a UnboundLocalError??

And do you want to handle it in the same way as the ValueError?

students/davidkan/session03/mailroom.py
```diff
@@ -0,0 +1,176 @@
...
+
+def add_donation_to_user(donor_name_index, donation_amount):
+    """
+    Function will add the amount given to the donation history of the selected user
+    :param donor_name_index: the index location of the donor
+    :param donation_amount: the amount the donor donated
+    :return: None
+    """
+    # Append the donation amount to the donor's name in the donor list data structure
+    donors[donor_name_index][1].append(donation_amount)
```
PythonCHB, line 80: if you got the donor object directly instead of the index, this could be:

`donor[1].append(donation_amount)`

students/davidkan/session03/mailroom.py
```diff
@@ -0,0 +1,176 @@
...
+    Function will print the thank you email to the donor on the terminal
+    :param full_name: username in lowercase
+    :param donation_amount: the amount the donor donated
+    :return: None
+    """
+    print()
+    # Finally, use string formatting to compose an email thanking the donor for their generous donation.
+    # Print the email to the terminal and return to the original prompt
+    print("Dear {0},\n\n"
+          "Thank you very much for your recent donation of ${1} to the local mailing room at our local charity.\n"
```
PythonCHB, line 94: nicely done!

students/davidkan/session03/mailroom.py
```diff
@@ -0,0 +1,176 @@
...
+    """
+    Function will send a thank you email and print it to the terminal
+    :return: None
+    """
+    while True:
+        resuts = input("\nFull Name:\n"
+                       "\n> ")
+        full_name = resuts.lower()
+
+        if 'list' in full_name:
```
PythonCHB, line 112: probably better to check for exact match to list, jsut in cse someone has a name that has "list" in it:

```
In [8]: 'list' in "Fred Plisternack"
Out[8]: True
```
OK, that's not a real name, but there may well be one like that....

students/davidkan/session03/mailroom.py
```diff
@@ -0,0 +1,176 @@
...
+            show_donor_list()
+            continue
+
+        if full_name not in [row[0] for row in donors]:
```
PythonCHB, line 116: we haven't learned this yet, but you can use a generator expression, rather than a list comprehension:

`if full_name not in (row[0] for row in donors):`

note the parens, rather than square brackets.

The difference is that the generator expression doesn't actually create a new list -- it will simply go thorugh it looking for the value, and then stop when it finds it.

a bit more efficient.

students/davidkan/session03/mailroom.py
```diff
@@ -0,0 +1,176 @@
...
+            add_donor(full_name)
+
+        donor_name_index = get_donors_index(full_name)
```
PythonCHB, line 119: this is where you could pass the donor object around, instead of keeping track of teh index:

```
donor = get_donor(full_name)
donation_amount = get_donation_amount()
add_donation_to_user(donor, donation_amount)
```
A little cleaner, and more extendable to a different data structure (like a dict...)

students/davidkan/session03/mailroom.py
```diff
@@ -0,0 +1,176 @@
...
+        donation_amount = get_donation_amount()
+        add_donation_to_user(donor_name_index, donation_amount)
+        print_email_to_terminal(full_name, donation_amount)
+
+        break
+
+
+def create_report():
+    """
+    Function will print out the report in a tabular formatted view
```
PythonCHB, line 129: nice doc strings!

... though you don't need to day "Function" -- it's clearly a function :)

students/davidkan/session03/mailroom.py
```diff
@@ -0,0 +1,176 @@
...
+    :return: None
+    """
+    # Create header for report
+    print("\n{:<10} | {:>10} | {:>10} | {:>10}".format('Donor Name', 'Total Given', 'Num Gift', 'Average Gift'))
+    print("{}".format(''.join(['-' * 52])))
+
+    # Create list that includes donor name, total given, number of gifts, and average gift as values in each row
+    donors_list = [(row[0], sum(row[1]), len(row[1]), int(sum(row[1]) / len(row[1]))) for row in donors]
```
PythonCHB, line 137: another nice list comp!

students/davidkan/session03/mailroom.py
```diff
@@ -0,0 +1,176 @@
...
+
+    # Sort total given by descending order
+    ordered_donors_list = sorted(donors_list, key=lambda x: (x[1],x[1]), reverse=True)
```
PythonCHB, line 140: very nice use of sorted() and lambda for the sort key.

but why (x[1],x[1]) ?? that's going to sort the exact same as x[1]

students/davidkan/session03/mailroom.py
```diff
@@ -0,0 +1,176 @@
...
+
+    # Iterate through the ordered donors list and print out the results
+    for items in ordered_donors_list:
+        print("{:<10} $ {:>11} $ {:>10} $ {:>11}".format(*items))
```
PythonCHB, line 144: perfect format string!

students/davidkan/session03/mailroom.py
```diff
@@ -0,0 +1,176 @@
...
+                  "\n> "))
+            if answer == 3:
+                break
+            elif answer == 1:
+                send_thank_you()
+            elif answer == 2:
+                create_report()
+            else:
+                print("\nPlease type 1, 2, or 3 only!\n")
+        except ValueError as v:
```
PythonCHB, line 169: no need for the "as v" if you aren't going to do anything with it.

PythonCHB (approved):
nice jop, good structure breaing things out into functions.

And I love the docstrings!

dkkan:
Thank you very much for the feedback. I learned a great deal from it.

PR #79, opened 2017-10-25
session 4 work

mattmaeda:


students/mattmaeda/session03/fruits.py
```diff
@@ -0,0 +1,14 @@
+def remove_fruit():
+    '''remove last or user specified fruit from list'''
+    fruits = ['Apples','Pears','Oranges','Peaches']
+    print(fruits)
+    del fruits[-1]
+    print(fruits)
+    choice = input("Which fruit should be removed? > ")
+    ## this works in ipython but not here...?
+    ## should remove all instances of choice
+    fruits2 = [x for x in fruits if x!= choice]
```
PythonCHB, line 10: nice list comprehension -- that should work fine.

note that the user will have needed to type the choice exactly the same way -- with no spaces or anything.

students/mattmaeda/session04/dict_lab.py
```diff
@@ -0,0 +1,48 @@
...
+print("Print values")
+print(master.values())
+
+print("#" * 5)
+print("Print if 'cake' is a key in the dictionary")
+print("cake" in master)
+
+print("#" * 5)
+print("Print if 'Mango' is a value  in the dictionary")
+print("Mango" in master.values())
```
PythonCHB, line 48: this all looks great!

students/mattmaeda/session04/file_lab.py
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env python3
+import os
+
+my_path = os.path.abspath(__file__)
```
PythonCHB, line 4: nice find!

usually, though you want to do stuff in relation to the current working directory --not the dir that that file is in.

Though running code from an IDE or editor can make that challenging...

students/mattmaeda/session04/file_lab.py
```diff
@@ -0,0 +1,22 @@
...
+languages = set()
+with open(text_file_path) as students:
+    for line in students:
+        try:
+            langs = line.strip().split(":")[1].replace(" ","").split(",")
+            if langs[0].strip()[0].isupper():
+                langs = langs[1:]
+            for lang in langs:
+                if len(lang) > 0:
+                    languages.add(lang)
```
PythonCHB, line 17: note that you can use `set.update()` to add a list of stuff all at once. might be easier than the loop.

students/mattmaeda/session04/file_lab.py
```diff
@@ -0,0 +1,22 @@
...
+        except IndexError:
```
PythonCHB, line 18: where might that index error have been raised? We haven't covered this yet -- so nice work! But you want to catch the exception as close to where it might be raised as possible. 

The contents of try blocks should be small...

students/mattmaeda/session04/mailroom_plus.py
```diff
@@ -0,0 +1,88 @@
+#!/usr/bin/env python3
+
+DONORS = {
```
PythonCHB, line 3: looks good.

students/mattmaeda/session04/trigrams.py
```diff
@@ -0,0 +1,63 @@
...
+    raw_list =  re.sub('[^0-9a-zA-Z]+', ' ', book_text).split(" ")
+    return [word.lower() for word in raw_list if word != ""]
+
+
+def build_word_dict(word_list):
+    word_dict = {}
+    for i in range(len(word_list)-2):
+        pair = tuple(word_list[i:i+2])
+        follower = word_list[i+2]
+        word_dict.setdefault(pair, []).append(follower)
```
PythonCHB, line 15: nice find!

students/mattmaeda/session04/trigrams.py
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env python3
+import re
+from random import randint
+
+def convert_to_word_list(book_text):
+    raw_list =  re.sub('[^0-9a-zA-Z]+', ' ', book_text).split(" ")
```
PythonCHB, line 6: you've gone beyond me on regexes -- nice!

students/mattmaeda/session04/trigrams.py
```diff
@@ -0,0 +1,63 @@
...
+    for i in range(len(word_list)-2):
+        pair = tuple(word_list[i:i+2])
+        follower = word_list[i+2]
+        word_dict.setdefault(pair, []).append(follower)
+
+    return word_dict
+
+
+def build_new_text(word_dict):
+    rand_idx = randint(0, len(word_dict.keys()))
```
PythonCHB, line 21: check out ``random.choice()``

PythonCHB, line 21: you can get the number of entries in a dict with `len(the_dict)` -- no need for the keys() call.

students/mattmaeda/session04/trigrams.py
```diff
@@ -0,0 +1,63 @@
...
+    new_list = []
+
+    idx = 0
+
+    start_follower_arr = None
+
+    for k,v in word_dict.items():
```
PythonCHB, line 28: this is pretty klunky --how about:

```
rand_key  = list(word_dict.keys())[rand_idx]
...
```
or better yet: `random.choice(list(word_dict.keys())`

students/mattmaeda/session04/trigrams.py
```diff
@@ -0,0 +1,63 @@
...
+        idx += 1
+
+    last_two = tuple(new_list[-2:])
+
+    while True:
+        try:
+            new_list = append_next_word(new_list, word_dict[last_two])
+            last_two = tuple(new_list[-2:])
+        except KeyError:
+            break
```
PythonCHB, line 43: This is going to happen if you get a pair that was not in the original text -- unlikely but possible.

But do you want to use it as your stopping criteria?

students/mattmaeda/session04/trigrams.py
```diff
@@ -0,0 +1,63 @@
...
+
+    return new_list
+
+
+def append_next_word(new_word_list, possible_words):
+    if len(possible_words) == 1:
+        new_word_list.append(possible_words[0])
```
PythonCHB, line 50: you can  generate a random number that can only be 0:

`random.randint(0,0)`

so no need for the special case.

or use random.choice again... handy function, eh?

students/mattmaeda/session04/trigrams.py
```diff
@@ -0,0 +1,63 @@
...
+
+    return new_word_list
+
+if __name__ == "__main__":
+    book_file = "sherlock_small.txt"
+    fh = open(book_file)
+    text = convert_to_word_list(fh.read())
+    word_dict = build_word_dict(text)
+    new_text = build_new_text(word_dict)
+    print(" ".join(new_text))
```
PythonCHB, line 63: Did you like what you ended up with?

PythonCHB (approved):
looks good -- all done after class Tuesday night? Wow!

PR #82, opened 2017-10-28
johnn session 04 exercises

johnnzz:
Re-add changes post conflict resolution by reverting and and re-applying new changes.


johnnzz:
EDIT:  Resolved conflicts.   

This PR originally had conflicts, probably because I squashed some commits and force pushed not realizing that some of those changes got piggy-backed and merged as part of the session03 pull request.

PR #83, opened 2017-10-28
session 4 mailroom.py submitted for review

ea8606:
added dicts and a function to write the donor thank you's to a file.


students/EricAdams/session04/mailroom.py
```diff
@@ -0,0 +1,201 @@
...
+donor_db = {"William Gates, III": [653772.32, 12.17],
+            "Jeff Bezos": [877.33],
+            "Paul Allen": [663.23, 43.87, 1.32],
+            "Mark Zuckerberg": [1663.23, 4300.87, 10432.0]}
+
+
+# loop through the donor list and print the 0th element of the list
+def print_donors():
+    """
+    loop through the donor list and print the 0th element of the list
```
PythonCHB, line 37: make sure to update the docstrings!

students/EricAdams/session04/mailroom.py
```diff
@@ -0,0 +1,201 @@
...
+    which is the donors name
+    donor name is now the key, since changing over to dict
+    """
+
+    print("Donor list:\n")
+    for donor in donor_db.keys():
```
PythonCHB, line 43: some as:

`for donor in donor_db:`

the simplest form uses the keys.

students/EricAdams/session04/mailroom.py
```diff
@@ -0,0 +1,201 @@
...
+def find_donor(name):
+    """
+    Find a donor in the donor db
+
+    :param: the name of the donor
+
+    :returns: The donor data structure -- None if not in the donor_db
+    """
+    donor = name.lower()
+    for key in donor_db.keys():
```
PythonCHB, line 56: so this kinda kills the point of using a dict. The great thing about a dict is fast lookup. But you're right, it needs to be an exact match. so you do need to do this loop if you maynot have an exact match.

Another solution -- use a "standardized" version of the name as the dict key (maybe lower(), spaces removed, or?

then store the "proper" version of the name in the values.

that will give you the quick lookup.

students/EricAdams/session04/mailroom.py
```diff
@@ -0,0 +1,201 @@
+#!/usr/bin/env python3
+
+"""
+Used Chris's solution to session 3 version and added to it to create this,
```
PythonCHB, line 4: fair enough -- but make sure you understand the code, and why i wrote it that way!

-CHB

PythonCHB (approved):
you got the key point, but do be careful when refactoring --with a different data structure, the logic may have to be different.

PR #84, opened 2017-10-29
Please review assignment and question about Git push

zandra2:
Please review session 04, mailroomDictFile.py assignment. 
Also, I am a little confused as to whether I still need to do a pull request on GitHub after I push a file from Git Bash. 

Thank you!


PythonCHB:
Hopefully I'll get to review this tomorrow.

But no, you don't need to make a new PR when you push new code, as long as the old PR is open.

You can always go on gitHub and look at the PR to make sure what you want is in there.

session04/mailroomDictFile.py
```diff
@@ -0,0 +1,126 @@
...
+    return question
+
+def print_donor_list():
+    """ printing 0th element from dict (printing only donors' name)"""
+    print("Donor Name:\n")
+    for name in donor_dict:
+        print(name)
+
+def find_donor(name):
+    """ finding donor's name in donor_dict and printing their information """
```
codedragon, line 39: Technically speaking, you are returning the donor's name, rather than printing anything here. Always a good idea to update your docstrings when you change your code. :)

session04/mailroomDictFile.py
```diff
@@ -0,0 +1,126 @@
...
+    for donor in donor_dict:
+        # case-insenstive compare
+        if name.lower() == donor.lower():
+            return donor
+    return None
```
codedragon, line 44: Not necessary to return None, python will do this automatically, if it doesn't find a donor.

session04/mailroomDictFile.py
```diff
@@ -0,0 +1,126 @@
...
+#     for name, amounts in donor_list.items():
+#         print("\nDear {},\n\nThank you for your kind donation of ${}."
+#           "\n\nSincerely,\n\nA Team".format(name, amounts))
+
+def letters_for_everyone():    #need to go back and put letters in txt file.
+    """ writing a thank you letter to all donor """
+    for name, amounts in donor_dict.items():
+        new_amounts = []
+        for i in amounts:
+            new_amounts.append('$' + str(i))
```
codedragon, line 67: You can change this to a list comprehension now!

session04/mailroomDictFile.py
```diff
@@ -0,0 +1,126 @@
...
+    else:
+        amounts = float(input("Enter donation amount "))
+        donor = find_donor(answer)
+        if donor is None:
+            donor_dict[answer] = [amounts]   # code on the left is correct (incorrect: donor = (answer, [float(amounts)]))
+        else:
+            donor_dict[donor].append(amounts)   # import pdb; pdb.set_trace()--checking to see why can't use append
+
+        # print a letter to thanks the donor
+        print(print_letter(answer, float(amounts)))
```
codedragon, line 108: Have you noticed that you are printing out None to the command line after the letter? This is because print will print whatever is returned by the function you are calling in it. In this case, print_letter is not actually returning anything, so it prints None. Print_letter is doing all of the printing within the function itself, so just a call to print_letter is sufficient to print the letter, no need to call print here.

session04/mailroomDictFile.py
```diff
@@ -0,0 +1,126 @@
...
+    """ printing a report that includes donor names, contribution amounts, sum, and average """
+    report = []
+    for (name, amount) in donor_dict.items():
+        total_amount = sum(amount)
+        number_gifts = len(amount)
+        avg_gift = total_amount / number_gifts
+        report.append((name, total_amount, number_gifts, avg_gift))
+
+    # sorting report data A-Z
+    report.sort(key=sort_key)
```
codedragon, line 84: It would be a good idea to split print_report into two functions: something like generate_report that returns the report variable, and the print stuff below this comment, that just prints stuff. That way you can write a test that makes sure the data in report is correct, without worrying about the report format.

codedragon (approved):
This is great! Works as advertised. :)

codedragon:
Hi, notice that your code is not in the students directory under your name. It looks like you have a directory there, please move this file there, and then submit a new pull request.

PR #85, opened 2017-10-29
trigrams.item()

HiroTestStudent:
Hi Chris and Maria,

Here are two files that I commited and pushed:

1. Trigrams: I added two approaches: 1. .item(), and 2. .setdefault(). For .setdefault(), I added it under the if statement (if pair not in trigrams: ). For some reason, else: clause shows error so I commented it out at this moment. The error message is "Traceback (most recent call last):
  File "trigrams_hiro.py", line 50, in <module>
    trigrams[pair].append(follower)        
AttributeError: 'str' object has no attribute 'append'"

2. Mailroom2: I converted the code from list to dict, with Chris's help during his coffee office hours today.

Thanks,
Hiro


students/hirot_directory/session04/mailroom2.py
```diff
@@ -0,0 +1,98 @@
+#!/usr/bin/env python3
+
+donor_names = ["William Gates III", "Mark Zuckerberg", "Jeff Bezos", "Paul Allen"]
+donations = [[200, 500], [2000, 3000], [4000, 5000], [100, 150]]
+donor_dict = dict(zip(donor_names, donations))
+
+def thank_you():
+
```
codedragon, line 8: This is a little clunky, and doesn't actually ever print a letter. What you really want here is another menu where you can either enter a name/donation (and print a letter!), or print 'list' to get all of the names (which returns to this same menu) or exits to the main menu).

students/hirot_directory/session04/trigrams_hiro.py
```diff
@@ -0,0 +1,51 @@
...
+trigrams = {}
+
+for i in range(len(words)-2):
+    pair = tuple(words[i:i+2])
+    follower = words[i+2]
+    print(pair, follower)
+    trigrams[pair] = follower
+
+print(trigrams.items())
+
```
codedragon, line 33: Hmm, since you are not using lists for your follower, you are losing some data. You have to replace that if block with something: Notice in the print statement from your dictionary above, "I wish" is followed once by "may" and once by 'I', but here you only get "I". The other change, printing as trigrams.items() rather than just printing the dictionary, really doesn't buy you much.

students/hirot_directory/session04/trigrams_hiro.py
```diff
@@ -0,0 +1,51 @@
...
+for i in range(len(words)-2):
+    pair = tuple(words[i:i+2])
+    follower = words[i+2]
+    print(pair, follower)
+    if pair not in trigrams:
+        trigrams.setdefault(pair, follower)        
+    # else:
+    #     trigrams[pair].append(follower)        
+print(trigrams)
```
codedragon, line 51: The purpose of setdefault is to initialize the value to something, if it is not there already, so you are close. However, you cannot append to a string (which is the error you were getting), so what you really want to do is setdefault to be a list, so you can append. So, your job is to figure out how setdefault would get rid of this conditional:
```
if pair not in trigrams:
    trigrams[pair] = [follower]
else:
    trigrams[pair].append(follower)
```
Note that is the same conditional you have in the first case, that we did in class. If you use setdefault to create a list (with nothing in it!), then you can append immediately, and get rid of your if/else statement. Here is an example: https://www.saltycrane.com/blog/2010/02/python-setdefault-example/

codedragon (approved):
You are making progress. Things to do (one at a time):
mailroom:  
* improve your thank_you function
* write thank you letters to disk

trigrams:
* get your dictionary working with set_default
* actually use the sherlock_small.txt text to make your dictionary
* use your dictionary to write something new

PR #87, opened 2017-10-29
Please review trigrams assignment

kmarguerite:
Requesting feedback on trigrams assignment.


students/kate/session04/mailroom2.py
```diff
@@ -0,0 +1,124 @@
...
+"""
+Mailroom Exercise, session 4
+using dictionaries
+"""
+import math
+
+# define global donor dict and $
+donors_dict = {
+    'Hermione Granger': [43.01, 13.17],
+    'Molly Weasley':87.03,
```
PythonCHB, line 12: the others are lists of donations, this is a single value -- hasn't that caused problems?

it should be a list with a single entry --then you don't need any special case code.

students/kate/session04/trigrams.py
```diff
@@ -0,0 +1,76 @@
...
+def gen_trigrams(words):
+    # make an empty dicitonary for the trigrams
+    trigrams_dict = {}
+    for i in range(len(words)-2):
+        pair = tuple(words[i:i+2])
+        follower = words[i+2]
+        print(pair, follower)
+        if pair in trigrams_dict:
+            trigrams_dict[pair].append(follower)
+        trigrams_dict[pair] = [follower]
```
PythonCHB, line 21: careful! this should be in an else block.

this way, you are always writing over the existing pair, even if it was there.

PythonCHB: oh, and you can use dict.setdefault to reduce it to one line.

students/kate/session04/trigrams.py
```diff
@@ -0,0 +1,76 @@
...
+    return(trigrams_dict)
+
+def build_new_text(word_dict):
+    # create index to hold a random integer
+    rand_index = randint(0, len(word_dict.keys()))
```
PythonCHB, line 26: this can be just `len(word_dict)`

students/kate/session04/trigrams.py
```diff
@@ -0,0 +1,76 @@
...
+    # create list to hold words
+    new_list = []
+    # set counter to 0
+    index = 0
+
+    # build sentences using index counter. Use for loop to add keys and values
+    for keys,values in word_dict.items():
+        if rand_index == index:
```
PythonCHB, line 34: this is pretty klunky -- you are looping through the entire dict until you get to the random index.

better to get it directly:

list(word_dict.keys())[rand_index]

or other options -- see:

https://github.com/UWPCE-PythonCert/IntroPython-2017/blob/master/solutions/Session04/arbitrary_key.py

(arbitrary_key.py in the solutions dir of the class repo)

students/kate/session04/trigrams.py
```diff
@@ -0,0 +1,76 @@
...
+            new_list.append(keys[0])
+            new_list.append(keys[1])
+            new_list = append_next_word(new_list, values)
```
PythonCHB, line 37: this is less than ideal -- you are passing the new_list in to be added to.

probably a little better to do the append here:

`new_list.append(get_next_word(values))`

That way it's more clear that the list is being appended to.

Also, maybe a better name than "values" -- maybe "followers" or something that describes what it actually means.

students/kate/session04/trigrams.py
```diff
@@ -0,0 +1,76 @@
...
+        try:
+            new_list = append_next_word(new_list, word_dict[last_two])
+            last_two = tuple(new_list[-2:])
+        except KeyError:
+            break
+
+    return new_list
+
+def append_next_word(new_word_list, words):
+    if len(words) == 1:
```
PythonCHB, line 53: no need to check for this special case -- randint(0,0) will always return zero...

students/kate/session04/trigrams.py
```diff
@@ -0,0 +1,76 @@
...
+        new_word_list.append(words[0])
+    else:
+        rand_index = randint(0, (len(words)-1))
+        new_word_list.append(words[rand_index])
```
PythonCHB, line 57: without the special case, maybe no need for a function at all. in the above:

`new_word_list.append(words[rand.randint(0, len(words))]))`

not a bad one-liner!

students/kate/session04/trigrams.py
```diff
@@ -0,0 +1,76 @@
...
+    return new_word_list
+
+# main function
+# use functions: make trigrams
+if __name__ == "__main__":
+    # open & read file, split file,
+    f = "sherlock_small.txt"
```
PythonCHB, line 64: you do want to at least go past the header....

PythonCHB (approved):
nice work.

A few tweaks needed, but you've got the logic down!

PR #90, opened 2017-10-30
Dictionaries in mail room !

guillaumethomas:


students/guillaume/session04/mailroom_dic.py
```diff
@@ -0,0 +1,186 @@
...
+from os import system
+
+
+def read_data_dic(textfile):
+    '''
+    Read data from a textfile about the donor and the
+    amount of donations
+    Pattern is as follow:
+    '''
+    file = open(textfile, 'r')
```
PythonCHB, line 11: use:

```
with open(testfile, 'r') as infile:
    ...
```

students/guillaume/session04/mailroom_dic.py
```diff
@@ -0,0 +1,186 @@
...
+    donors = list()
+    for i in list(file.readlines()):
```
PythonCHB, line 13:
```
for i in infile:
   ...
```

PythonCHB (approved):
we did this in class, so I'll just merge.

PR #93, opened 2017-10-31
Some session 4 codes

chops77:


students/Chao/session04/dict_lab.py
```diff
@@ -0,0 +1,88 @@
...
+    
+    # Build empty sets
+    s2 = set()
+    s3 = set()
+    s4 = set()
+
+    # Create sets s2, s3 and s4 that contain numbers from zero through twenty, divisible 2, 3 and 4
+    for n in range(21):
+        if n%2 == 0:
+            s2.update([n])
```
PythonCHB, line 51: `.update()`  is for adding a bunch of stuff all at once (why you needed the []). If you just want to add one thing, you can use `.add()`

students/Chao/session04/file_lab.py
```diff
@@ -0,0 +1,61 @@
...
+        # Check if item is a file (not a directory)
+        if os.path.isfile(f):
+            print(f)
+
+def copyf(inputf, outputf):
+    """ Write a program which copies a file from a source, to a destination """
+
+    with open(inputf,'rb') as f_in, open(outputf,'wb') as f_out:
+        while True:
+            # Read to buffer, 1 byte at a time
```
PythonCHB, line 18: a 1 byte buffer is pretty darn small!

maybe 1k at a time? -- 1024 bytes?

students/Chao/session04/mailroom2.py
```diff
@@ -0,0 +1,88 @@
+#!/usr/bin/env python
```
PythonCHB, line 1: this all looks great!

PythonCHB (commented):
not sure what the spongebob image is for -- but in any case, you don't want the backup in the repo as well.

I highly recommend that you add files one by one, so as not to add stuff you don't intents to add.

students/Chao/session04/students_backup.txt
```diff
@@ -0,0 +1,31 @@
+Name: Languages
```
PythonCHB, line 1: you don't want to save a backup in git. git IS a backup :-)

PythonCHB:
I look  forward to your trigrams solution....

PR #94, opened 2017-10-31
adding mailroom for feedback

briannagata:


students/briannagata/session03/mailroom/actions.py
```diff
@@ -0,0 +1,30 @@
+# Modules to process selections made in mailroom.py
+# Python 3.6
+
+
+def list_database(donors):
+    """ Print the list of donor names"""    
+    for name in donors.keys():
```
briannagata, line 7:
```
for name in donors:
...
```

codedragon: Yep, for accessing the keys you actually don't need to call keys(). The keys() function is useful for providing a set-like object, which you could use for comparing the keys of different dictionaries, for example.

students/briannagata/session03/mailroom/actions.py
```diff
@@ -0,0 +1,30 @@
...
+        print(name)
+
+
+def send_thankyou(name, money, text):
+    """ Emulate a thank you email by emulating content to terminal. """
+    text = text.replace('[DONOR]', name)
+    text = text.replace('[DONATION]', '{:.2f}'.format(money))
+    print(text, end='\n')
+
```
codedragon, line 16: Clever way of making use of a thank you letter template. This is great, since it completely separates the letter formatting from the data, making testing easier. All you have to do is make sure you are sending the correct name and money in.

students/briannagata/session03/mailroom/actions.py
```diff
@@ -0,0 +1,30 @@
...
+                                             'Total Given',
+                                             'Num Gifts',
+                                             'Average Gift'))
+    print('-' * 63)
+
+    for key, value in sorted(donors.items(), key=lambda x: x[1], reverse=True):
+        print('{:20} ${:10.2f} {:13}   ${:11.2f}'.format(key,
+                                                         sum(value),
+                                                         len(value),
+                                                         sum(value)/2))
```
codedragon, line 30: Since you are sorting the dictionary on the first item of the values, you will get things sorted by the first donation, rather than the sum.

students/briannagata/session03/mailroom/mailroom.py
```diff
@@ -0,0 +1,75 @@
...
+    elif response == '3':
+        for donor, donation in donors.items():
+            print()
+            actions.send_thankyou(donor,
+                                  sum(donation),
+                                  e_template)
+
+    # Exit
+    elif response == '4':
+        raise SystemExit
```
codedragon, line 75: This is a very long conditional. How could you condense it?

codedragon:
Nice solution. Think about how you might use a dictionary to make your conditionals cleaner.

codedragon (approved):
Nice solution. Think about how you might use a dictionary to make your conditionals cleaner.

students/briannagata/session03/mailroom/actions.py
briannagata: updated code to:
```
for key, value in sorted(donors.items(), key=lambda x: sum(x[1]), reverse=True):
    # ...
```

students/briannagata/session03/mailroom/actions.py
briannagata: Took liberty to find a way to avoid repeating myself:
```
def send_thankyou(name, money, text):
    """ Emulate a thank you email by emulating content to terminal. """
    replacements = {'[DONOR]': name,
                    '[DONATION]': '{:.2f}'.format(money)}

    for key, value in replacements.items():
        text = text.replace(key, value)

    print(text, end='\n')
```

PR #97, opened 2017-11-01
comprehension and exception problems

mattmaeda:
add exception handling and use of comprehensions to mailroom project

### DIFF
--- a/students/mattmaeda/session05/comprehensions.py
+++ b/students/mattmaeda/session05/comprehensions.py
@@ -1,0 +1,105 @@
+#!/usr/bin/env python3
+
+# Problems from
+# https://uwpce-pythoncert.github.io/PythonCertDevel/exercises/comprehensions_lab.html#exercise-comprehensions
+
+def count_evens(num_list):
+    # http://codingbat.com/prob/p189616
+    # Using a list comprehension, return the number of even integers in the
+    # given list.
+    return len([i for i in num_list if i % 2 == 0])
+
+
+def dict_formatter(ref_dict):
+    # dict/set problem 1
+    #
+    # Print the dict by passing it to a string format method, so that you get
+    # something like:
+    #
+    # “Chris is from Seattle, and he likes chocolate cake, mango fruit, greek
+    # salad, and lasagna pasta”
+    return "%(name)s is from %(city)s, and he likes %(cake)s cake, %(fruit)s " \
+           "fruit, %(salad)s salad and %(pasta)s pasta" % {k: v for k, v in ref_dict.items()}
+
+
+def int_to_hex_mapping(n):
+    # dict/set problems 2 and 3
+    #
+    # Using a list comprehension, build a dictionary of numbers from zero to
+    # fifteen and the hexadecimal equivalent (string is fine). (the hex()
+    # function gives you the hexidecimal representation of a number as a string)
+    #
+    # Do the previous entirely with a dict comprehension – should be a one-liner
+    return {i: hex(i)for i in range(n+1)}
+
+
+def count_letter_a_in_dict(ref_dict):
+    # dict/set problem 4
+    #
+    # Using the dictionary from item (1): Make a dictionary using the same keys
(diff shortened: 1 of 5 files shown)


students/mattmaeda/session05/comprehensions.py
```diff
@@ -0,0 +1,105 @@
...
+    # dict/set problem 1
+    #
+    # Print the dict by passing it to a string format method, so that you get
+    # something like:
+    #
+    # “Chris is from Seattle, and he likes chocolate cake, mango fruit, greek
+    # salad, and lasagna pasta”
+    return "%(name)s is from %(city)s, and he likes %(cake)s cake, %(fruit)s " \
+           "fruit, %(salad)s salad and %(pasta)s pasta" % {k: v for k, v in ref_dict.items()}
+
```
codedragon, line 23: Note there is a way to do this, so that you do not have to write a dictionary comprehension for formatting. To do this you will have to use the format method instead of %.

students/mattmaeda/session05/comprehensions.py
```diff
@@ -0,0 +1,105 @@
...
+    #          code. you will end up with a list of sets – one set for each
+    #          divisor in your sequence.
+    #       3. The idea here is that when you see three (Or more!) lines of code
+    #          that are almost identical, then you you want to find a way to
+    #          generalize that code and have it act on a set of inputs, so the
+    #          actual code is only written once.
+    #   c. Extra credit: do it all as a one-liner by nesting a set comprehension
+    #      inside a list comprehension. (OK, that may be getting carried away!)
+    return [{i for i in range(max_range+1) if i % d == 0} for d in divisors]
+
```
codedragon, line 66: Great work! Note that for integers, 0 is False and other numbers are True, so you can replace `if i % d == 0` with `if not i % d` Not much shorter, nor necessarily more readable, but it does get used, so worth being familiar with. I think computationally the if not idiom is slightly faster, but rarely would that matter.

students/mattmaeda/session05/comprehensions.py
```diff
@@ -0,0 +1,105 @@
...
+
+    assert count_letter_a_in_dict(food_prefs) == {"name": 0, "city": 1,
+                                                  "cake": 1, "fruit": 1,
+                                                  "salad": 0, "pasta": 3}
+
+    assert set_comprehensions([2,3,4], 20) == [{0, 2, 4, 6, 8, 10, 12, 14, 16, 18, 20},
+                                               {0, 3, 6, 9, 12, 15, 18},
+                                               {0, 4, 8, 12, 16, 20}]
+    print("All tests passed!")
```
codedragon, line 105: Yay! Nice tests! Looks like it is time to start writing unit tests in their own file, since your if main block is getting pretty large.

students/mattmaeda/session05/except_exercise.py
```diff
@@ -0,0 +1,54 @@
...
+# parameters. (pun intended)
+
+langs = ['java', 'c', 'python']
+
+try:
+    more_joke = more_fun(langs[1])
+except IndexError:
+    fun(more_joke)
+else:
+    last_fun()
```
codedragon, line 54: I fear my instructions for this last one were not clear enough. I'll have to work on that. Note that else runs "if nothing else goes wrong", so to run something always, regardless of whether there was an error, you need a different key word.

students/mattmaeda/session05/mailroom_plus_with_exceptions.py
```diff
@@ -0,0 +1,91 @@
...
+
+            elif choice == 1:
+                mailroom.thanks()
+            elif choice == 2:
+                mailroom.report()
+            else:
+                quit = True
+
+        except ValueError:
+            mailroom.invalid_choice(response)
```
codedragon, line 91: Note that you should include only the code that you expect to fail in the try/except block, otherwise you may catch exceptions you did not expect to catch. In this case, what you want to catch would be the int conversion.

codedragon (approved):
This looks very nice!

PR #98, opened 2017-11-02
Kata_test Review

brhelodrvr:
@PythonCHB, I'm only interested in having `kata_test.py` reviewed, but can't figure out how to limit this pull request to only chose it.  In this 'Open a pull request' section, I see that I have 14 commits with 6 files that have changed.  I believe that you'll see all of those files come across.


PythonCHB:
@brhelodrvr: the way to limit a PR to just a few files is to make a branch in git, and put only those files in the branch. But that's getting deeper into git than we expect you to go now, so it's fine to jsut put the whole pile in a PR, and tell us which parts you'd like us to look at -- like you just did.

students/brian_warn/session04/kata_test.py
```diff
@@ -0,0 +1,94 @@
+#!/Library/Frameworks/Python.framework/Versions/3.6/bin/python
+# Kata Fourteen
+import random
+
+
+def main():
+    exceptions = ['.', '?', ' ', '!', ',', '-']
+
+    def build_trigram(infile):
```
PythonCHB, line 9: you do'nt want to define these functions inside the main() function --they should all be at the top level, and then you can call them from main().

students/brian_warn/session04/kata_test.py
```diff
@@ -0,0 +1,94 @@
...
+        ''' Function that builds the trigram dictionary from the source file. '''
+
+        trigrams = {}
+        with open(infile) as sherlock:
+            sherlock.readline()
+            for line in sherlock:
+                line = line.rstrip('\r\n')
```
PythonCHB, line 16: .strip(), .rstrip(), etc will strip any whitespace by default -- so better not to specify \r\n, jsut in case there are spaces or tabs, or ??? in there too.

students/brian_warn/session04/kata_test.py
```diff
@@ -0,0 +1,94 @@
...
+                for i in range(len(word) - 2):
+                    pair = tuple(word[i:i + 2])
+                    follower = word[i + 2]
+                    if follower not in exceptions:
+                        pass
+                    # Find dictionary method that does this in one call
+                    if pair in trigrams:
+                        trigrams[pair].append(follower)
+                    if pair not in trigrams:
+                        trigrams[tuple(pair)] = [follower]
```
PythonCHB, line 27: pair is already a tuple.

students/brian_warn/session04/kata_test.py
```diff
@@ -0,0 +1,94 @@
...
+        with open(infile) as sherlock:
+            sherlock.readline()
+            for line in sherlock:
+                line = line.rstrip('\r\n')
+                word = line.split(" ")
+                for i in range(len(word) - 2):
+                    pair = tuple(word[i:i + 2])
+                    follower = word[i + 2]
+                    if follower not in exceptions:
+                        pass
```
PythonCHB, line 22: this does nothing at all -- is that because you are holding the place until you decide what to put there?

or did yo intend to put "continue" in there, which would skip the rest of the code and continue the loop at the next pair...

careful with that though, the "exeption" would then show up in the next pair...

I'm not totally sure where you are heading here but if you want to ignore the punctuation, better to removeit from the text before running this loop.

students/brian_warn/session04/kata_test.py
```diff
@@ -0,0 +1,94 @@
...
+                    if pair in trigrams:
+                        trigrams[pair].append(follower)
+                    if pair not in trigrams:
+                        trigrams[tuple(pair)] = [follower]
+
+        # Check to see what's in trigrams dictionary:
+        # items = trigrams.items()
+        # for x in items:
+        #     print(x)
+        return trigrams.items()
```
PythonCHB, line 33: I'm pretty sure you want to return the entire trigram dict....

students/brian_warn/session04/kata_test.py
```diff
@@ -0,0 +1,94 @@
...
+
+    def get_trigram_key(trigram_items, upper):
+        ''' Generates a two-word tuple. '''
+
+        # print(trigram_items)
+        len_random_choice = 1
+        # Needed since some of the random_choice results may only have one word in them.
```
PythonCHB, line 40: huh?

I think I see what you are doing, but you're making it more complicated that it has to be.

the trigrams dict you created above is a gooddata structure for building new text, too.

the keys are word pairs, and the values are lists of the following words...

So to start it off, you want a random word pair, that is a random key from the dict:

`start_pair = random.choice(list(trigrams.keys()))`

you need the list() call there 'cause random.choice does not work with a dict keys object.

now you have a random pair, you can start bulding text by randomly choosinf one ofits followingwords:

`next_word = random.choice(trigrams[start_pair]))`

and away you go....

students/brian_warn/session04/kata_test.py
```diff
@@ -0,0 +1,94 @@
...
+        tg_key_start = get_trigram_key(trigram_items, 1)
+        print("trigram key is now: ", tg_key_start)
+        # Rest of sentence (i.e. upper -> 0)
+        # TODO: Encapsulate in a while loop that defines when the sentence-building process is finished
+        # TODO: Define the criteria that specifies how long the loop will run.
+        tg_key_remaining = get_trigram_key(trigram_items, 0)
+        print("rest-of-sentence trigram is: ", tg_key_remaining)
+
+
+    def run_program():
```
PythonCHB, line 80: this should be your main() function, right there.

PythonCHB (approved):
Good start -- a little cleaning up and you'll have it!

students/brian_warn/session04/kata_test.py
brhelodrvr: So, if I follow you correctly here `def main():` will be just above my `if __name__ == "__main__":` section and `def main():` will contain the code that I currently stuff into `def run_program():`.  I then won't need to call `run_program` as I've been doing.  Is my thinking correct on this point?

students/brian_warn/session04/kata_test.py
brhelodrvr: Makes sense now, @PythonCHB.

students/brian_warn/session04/kata_test.py
brhelodrvr: I was still sorting out what to do here.  I think your `setdefault` approach is the direction I need to take.

students/brian_warn/session04/kata_test.py
PythonCHB: yup --that's it.

The basic flow is:

define all the functions (and global object)

define a main() function that calls the functions in the right order, etc...

call main() in the if __name__ == "__main__": block.

students/brian_warn/session04/kata_test.py
brhelodrvr: This shows an area where I hadn't worked through the next step of the sentence-building task.  My thought at this point was that I just needed the key in order to go through `build_trigram` again to get the next word.

students/brian_warn/session04/kata_test.py
brhelodrvr: Good.  This confirms my `run_program()` vs `main()` response I noted earlier.

students/brian_warn/session04/kata_test.py
brhelodrvr: Agreed that this is complicated.  I'll work with your recommendation and see how my code needs to change accordingly

As to where I was coming from, my approach generated the two scenarios below.  The short explanation is that I would produce results where I didn't get two words in the pair (the second run).  The first run shows that I got the two-word key that I wanted.

First Run:
```
raphael:session04 brwarn$ ./kata_test.py 
trigram_items type:  <class 'dict_items'>
random_choice in def is:  ('I', 'had')
random_choice length is:  2
trigram key is now:  ('I', 'had')
random_choice in def is:  ('Baker', 'Street.')
random_choice length is:  2
rest-of-sentence trigram is:  ('Baker', 'Street.')
```

Second Run:
```
raphael:session04 brwarn$ ./kata_test.py 
trigram_items type:  <class 'dict_items'>
random_choice in def is:  ['pacing', 'at']
random_choice length is:  2
trigram key is now:  ['pacing', 'at']
random_choice in def is:  ['now']
random_choice length is:  1
random_choice in def is:  ['associated']
random_choice length is:  1
random_choice in def is:  ('incidents', 'of')
random_choice length is:  2
rest-of-sentence trigram is:  ('incidents', 'of')
```

students/brian_warn/session04/kata_test.py
PythonCHB: What .items() does is return both the keys and the values as an iterable (kind of like a list). So you have all the information, but have lost the ability to efficiently look up a particular word pair.

And looking up a word pair is exactly what you need to to to build a new text.

PR #101, opened 2017-11-04
session5 work

Eowyn42:
I'm seeking feedback on the mailroom exercise. Specifically, when I implemeted dicts to handle the "cases" of different kinds of responses for the thank_you and make_report tasks, I ran into an issue returning to the main menu.  I'm trying to call the function from within a dict loop: arg_dict = {"1": update_donor, "2": list_donors, "3": return_to_menu}  then arg_dict.get(answer)(). return_to_menu CANNOT break the loop directly, that's a syntax violation. It can't just modify a variable within it that is checked one level up; that's a scope issue, which I've considered using a global variable to get around but I dislike that option. It could return a boolean value, which is checked one level up; then the other little helper functions would also all need to return a boolean. Or it could return an exception, which is handled one level up by breaking the loop -- That is what I have currently implemented, but I dislike using exceptions in this way. 

There must be an idiomatic way to do this. I got the try-catch idea from stack overflow; they suggested writing  a custom exception for the occasion, but I skipped that temporarily. Maybe I need to handle the return_to_menu user choice separately from everything else. That might be cleanest. I'm not sure. Thank you.

### DIFF
--- a/students/eowyn/session05/mailroom.py
+++ b/students/eowyn/session05/mailroom.py
@@ -1,0 +1,161 @@
+#!/usr/bin/env python
+
+import sys
+
+''' global variables'''
+donor_names = ["Margaret Atwood", "Fred Armisen",
+               "Heinz the Baron Krauss von Espy"]
+donations = [[300, 555], [240, 422, 1000], [1500, 2300]]
+DONORS = dict(zip(donor_names, donations))
+
+
+
+def safe_input():
+    return None
+
+def quit_code():
+    sys.exit()
+
+def return_to_menu():
+    ''' Raise a zero division error to trigger exit out of loop'''
+    return 5/0
+
+def list_donors():
+    ''' List all DONORS '''
+    print("All Donors:")
+    [print(x) for x in DONORS]
+
+
+def generate_table():
+    ''' Pretty-print a table of donor statistics '''
+    [name, total, number, ave] = [[], [], [], []]
+    for x in DONORS:  # loop over the keys which are donor names
+        name.append(x)
+        total.append(round(sum(DONORS[x])))
+        number.append(len(DONORS[x]))
+        ave.append(round(total[-1] / number[-1]))
+    report_data = list(zip(name, total, number, ave))
+    report_data = sorted(
+        report_data, key=lambda y: int(y[1]), reverse=True)
+    print_table(report_data)
+
+
+def send_letters():
+    ''' Send thank you notes to everyone in the DONORS dict'''
+    for donor in DONORS:
+        outfile = donor.replace(" ", "_") + '.txt'
+        with open(outfile, 'w') as f:
+            f.write(generate_letter(donor))
+    print("Successfully saved letters for each donor.")
+
+
+def generate_letter(donor_name):
+    '''
+    Generate a formatted thank you note to a donor and return
+    the string of the thank you note for printing or writing.
+    '''
+    fs = "Thank you, {}, for your generosity and recent gift of ${}."
+    return fs.format(donor_name, DONORS[donor_name][-1])
+
+
+def remove_inputquotes(a_string):
+    '''Remove auxillary quotes to cleanse a_string'''
+    a_string.replace('"', '')
+    a_string.replace('"', '')
+    return a_string
+
+
+def print_table(list_data):
+    ''' Pretty-print the donor records '''
+    headers = ('Donor Name', 'Total Given', 'Num Gifts', 'Average Gift')
+    fs1 = '|'.join(["{:<40}", "{:<12}", "{:<10}", "{:<12}"])
+    width = len(fs1.format(*headers))
+    fs2 = ''.join(["{:<40}", "${:^12.2f}", "{:^12d}", "${:^12.2f}"])
+    print(fs1.format(*headers))
+    print(width * "-")
+    for i in range(len(list_data)):
+        entry = list_data[i]
+        print(fs2.format(*entry))
+
+def update_donor():
+    '''
+    Add new donation to DONORS; if new donor, add to DONOR.
+    Then print a thank-you note for the donation.
+    '''
+    fullname = input("Enter a donor name (new or existing):\n")
+    fullname = remove_inputquotes(fullname)
+    if fullname in DONORS:
+        print("Existing Donor")
+        DONORS[fullname].append(float(input("Donation amount: ")))
+        print(generate_letter(fullname))
+    else:
+        print("New Donor")
+        DONORS[fullname] = [float(input("Donation amount: "))]
+        print(generate_letter(fullname))
+
+def thank_you():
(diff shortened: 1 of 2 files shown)


PythonCHB:
Agreed -- don't go the global route!

I'll take a look at what you've got now.

students/eowyn/session05/mailroom.py
```diff
@@ -0,0 +1,161 @@
...
+
+def safe_input():
+    return None
+
+def quit_code():
+    sys.exit()
+
+def return_to_menu():
+    ''' Raise a zero division error to trigger exit out of loop'''
+    return 5/0
```
PythonCHB, line 21: if you want to use Exception Handling for this, which is a fine idea, then you should specifically raise an Error:

raise ValueError

for instance, rather than using random code that happens to trigger an error.

As you mentioned, a custom Exception may be ideal, but you should be able to find a built in one that makes more sense than a ZeroDivisionError.

try:

`print([ er for er in dir(__builtins__) if "Error" in er])`

To see all the built in errors. Maybe SystemError would work? or ValueError?

students/eowyn/session05/mailroom.py
```diff
@@ -0,0 +1,161 @@
...
+    the string of the thank you note for printing or writing.
+    '''
+    fs = "Thank you, {}, for your generosity and recent gift of ${}."
+    return fs.format(donor_name, DONORS[donor_name][-1])
+
+
+def remove_inputquotes(a_string):
+    '''Remove auxillary quotes to cleanse a_string'''
+    a_string.replace('"', '')
+    a_string.replace('"', '')
```
PythonCHB, line 64: why call it twice? it will replace all occurances with one call.

and this isn't doing what you want -- strings are immutable, so it's returning a new string with the replacement, but then throwing it away. you want:

a_string = a_string.replace('"', '')

and you may want to call strip() as well to get rid of extra whiltespace...

students/eowyn/session05/mailroom.py
```diff
@@ -0,0 +1,161 @@
...
+    arg_dict = {"1": update_donor, "2": list_donors, "3": return_to_menu}
+    try:
+        while True:
+            try:
+                answer = input("To send a thank you, select one:\n "
+                               "(1) Update donor and send thank-you\n"
+                               "(2) List all existing DONORS\n"
+                               "(3) Return to main menu\n >")
+            except (EOFError, KeyboardInterrupt, TypeError):
+                safe_input()
```
PythonCHB, line 107: if you are going to use safe_input() here (which may not be  best --that was jsut an Exercise!) -- then you use it in place of input() -- rather than only if there is an Error.

students/eowyn/session05/mailroom.py
```diff
@@ -0,0 +1,161 @@
...
+            else:
+                try:
+                    arg_dict.get(answer)()
+                except (KeyError, TypeError):
```
PythonCHB, line 111: you are using strings as your keys, so you shouldn't get a TypeError anywhere.

students/eowyn/session05/mailroom.py
```diff
@@ -0,0 +1,161 @@
...
+
+''' global variables'''
+donor_names = ["Margaret Atwood", "Fred Armisen",
+               "Heinz the Baron Krauss von Espy"]
+donations = [[300, 555], [240, 422, 1000], [1500, 2300]]
+DONORS = dict(zip(donor_names, donations))
+
+
+
+def safe_input():
```
PythonCHB, line 13: if you're going to use this --the checks for the Exceptions should be in this function -- then you can jsut replace "input" with "safe_input" everywhere you want to use it.

-CHB

PythonCHB (commented):
Nice work -- you've almost got it -- and good job googling to find solutions!

students/eowyn/session05/mailroom.py
```diff
@@ -0,0 +1,161 @@
...
+def list_donors():
+    ''' List all DONORS '''
+    print("All Donors:")
+    [print(x) for x in DONORS]
+
+
+def generate_table():
+    ''' Pretty-print a table of donor statistics '''
+    [name, total, number, ave] = [[], [], [], []]
+    for x in DONORS:  # loop over the keys which are donor names
```
PythonCHB, line 32: good idea to use meaningful names -- `for name in DONORS` maybe?

students/eowyn/session05/mailroom.py
```diff
@@ -0,0 +1,161 @@
...
+
+def list_donors():
+    ''' List all DONORS '''
+    print("All Donors:")
+    [print(x) for x in DONORS]
+
+
+def generate_table():
+    ''' Pretty-print a table of donor statistics '''
+    [name, total, number, ave] = [[], [], [], []]
```
PythonCHB, line 31: no need for the square brackets here:

` name, total, number, ave = [], [], [], []`

does the same thing, slightly more efficiently.

students/eowyn/session05/mailroom.py
```diff
@@ -0,0 +1,161 @@
...
+    for x in DONORS:  # loop over the keys which are donor names
+        name.append(x)
+        total.append(round(sum(DONORS[x])))
+        number.append(len(DONORS[x]))
+        ave.append(round(total[-1] / number[-1]))
+    report_data = list(zip(name, total, number, ave))
+    report_data = sorted(
+        report_data, key=lambda y: int(y[1]), reverse=True)
```
PythonCHB, line 39: I think y[1] will already be an int...

students/eowyn/session05/mailroom.py
```diff
@@ -0,0 +1,161 @@
...
+    a_string.replace('"', '')
+    a_string.replace('"', '')
+    return a_string
+
+
+def print_table(list_data):
+    ''' Pretty-print the donor records '''
+    headers = ('Donor Name', 'Total Given', 'Num Gifts', 'Average Gift')
+    fs1 = '|'.join(["{:<40}", "{:<12}", "{:<10}", "{:<12}"])
+    width = len(fs1.format(*headers))
```
PythonCHB, line 72: if you are only going to use it once, you might as well simply hard-code the header

students/eowyn/session05/mailroom.py
```diff
@@ -0,0 +1,161 @@
...
+                    arg_dict.get(answer)()
+                except (KeyError, TypeError):
+                # errors are for if key doesn't exist or is None
+                    continue
+    except ZeroDivisionError:
+        pass
+
+def print_report():
+    ''' Print donor report or return to menu '''
+    arg_dict = {"1": generate_table, "2": return_to_menu}
```
PythonCHB, line 119: if those are the only two options -- the dict-based selection is probably overkill...

PythonCHB (approved):
some little things to tweak, but nice work!

PR #102, opened 2017-11-04
Session 05 assignment

HiroTestStudent:
Hi Chris and Maria,

The session 05 assignment.

1.  Mailroom5.py: I updated mailroom script. Please discard mailroom3. 

2. Trigrams_sherlock_small.py: I tried to strip ".", "," and "--" when the script read file but it shows the list doesn't have the strip function so I couldn't strip...

Thanks,
Hiro


students/hirot_directory/session04/trigrams_sherlock_small.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+def read_file():
+
+    lines = []
+
+    with open('sherlock_small.txt', 'r') as f:
+        while True:
+            line = f.readline().split()
+            lines.extend(line)
```
codedragon, line 17: The split is where the string is being turned into a list, so if you want to strip stuff first, you need to read the line, strip stuff and then split. A good exercise for you would be to do this in 3 lines, one reading, one stripping, and one splitting, and then figure out how to combine those three lines to do it in one line.

students/hirot_directory/session04/trigrams_sherlock_small.py
```diff
@@ -0,0 +1,72 @@
...
+            if not line:
+                break
+            # print(line)
```
codedragon, line 20: Note using `if not line` to stop your loop means that the loop will stop at the first blank line. This is probably not what you want. We were able to use this syntax to read in whole files earlier, because we were not processing the line before testing it. Once you use split, you are getting rid of the command keys (end of line characters), so there is no longer any difference between an empty line and the final line. You may have to look at the text file to see if there is something else you can use as a stopping point. There is an obvious stopping point in the big file, but if you are using the small one, you may have to get more creative, or check for the end before processing the line, but this adds more complexity.

students/hirot_directory/session04/trigrams_sherlock_small.py
```diff
@@ -0,0 +1,72 @@
...
+    for d in range(20):
+        pair = tuple(paragraph_list[-2:])
+        # pair = tuple(pair)
+        new_word = random.choice(tri_dict[pair])
+        paragraph_list.append(new_word)
+        # new_tuple = (i[1], new_word)
+        # i = new_tuple
+        # print(paragraph_list)
+    print(" ".join(paragraph_list))
+
```
codedragon, line 57: Since you are using just the first line of the text file, you will run into a pair that is not in your dictionary pretty quickly, definitely before you hit 20. Even once you fix your code so you are getting a larger dictionary, you may want to put a try/except in your loop, so you exit gracefully once you hit the last two words in the file (which will not be in the dictionary, since they do not have a word following them).

students/hirot_directory/session05/mailroom5.py
```diff
@@ -0,0 +1,147 @@
...
+                                           msg3="Num Gifts",
+
+                                           msg4="Average Gift"))
+    for d in donor_dict:
+        t = sum(donor_dict[d])
+        n = len(donor_dict[d])  # list.count() is to count inside of the list (i.e. # by blue)
+        a = t / n
+
+        print("{d: <50} ${t: 12.2f}{n: 12d}{a: 14.2f}".format(d=d, t=t, n=n, a=a))
+
```
codedragon, line 36: Looking good! See if you can figure out how to print them in order of total contributions.

students/hirot_directory/session05/mailroom5.py
```diff
@@ -0,0 +1,147 @@
...
+def make_donation():
+    donor_name = str(input("Full Name: "))
+    new_donation = int(input("How much do you want to donate?: "))
+
+    if (donor_name not in donor_dict):
+        donor_dict[donor_name] = []
+
+    donor_dict[donor_name].append(new_donation)
+    print("Thank you for donation, Mr./Ms. {}! Your donation amount is USD ${}.".format(donor_name, new_donation))
```
codedragon, line 46: Awesome, we are writing a letter now. Next step would be to make sure that you don't end up with two different entries in your dictionary for the same person, because of capitalization!

students/hirot_directory/session05/mailroom5.py
```diff
@@ -0,0 +1,147 @@
...
+                          ))
+            dict_choice[choice]()
+
+        except ValueError as e:
+            print("Error happened. Error is ", type(e) , "Please type 1, 2 or 3")
+        except KeyError as k:
+            print("Error happened. Error is ", type(k) , "Please type 1, 2 or 3")
+        # finally:
+        #   return mainloop()
+
```
codedragon, line 85: Nicely done! And good solution for returning to the main menu.

students/hirot_directory/session05/mailroom5.py
```diff
@@ -0,0 +1,147 @@
...
+            f = open('{0}.txt'.format(name), 'wb')
+            d = sum(d)
+            text = "Dear Mr./Ms. {},\n\n    Thank you for your kind donation of USD ${}.\n    It will be put to very good use.\n\n          Sincerely,\n            -The Team".format(n, d).encode()
+            f.write(text)
+            f.close()
+
+    except IOError as i:
+        print("Error code is", i)
+    print("letter drafts are in the folder for your review. Please review before mailing out.")
+
```
codedragon, line 121: Note that if there is an error code, there may not be letters for review, so you may want to print that only if nothing goes wrong. Hint: you can use else with exceptions: I think of it as "else if nothing goes wrong"

codedragon (approved):
You are making lots of great progress. I know it is hard when you feel others are farther along than you, but you are getting there. Keep writing code, and it will get easier.

students/hirot_directory/session04/trigrams_sherlock_small.py
codedragon: You know, if you use the `for line in lines:` idiom instead of a while loop, then you will just get all of the lines in  the file, and you don't have to worry about figuring out the end of the file. That being said, for the big file, there is a bunch of stuff at the end that you don't want, so you may still prefer to use custom endpoint.

PR #103, opened 2017-11-05
trigrams.py

ea8606:
Issues with output repeating words


students/EricAdams/session04/trigrams.py
```diff
@@ -1,16 +1,101 @@
...
-    pair = tuple(words[i:i+2])
-    follower = words[i+2]
-    print(pair, follower)
-    if pair in trigrams:
-        trigrams[pair].append(follower)
+import random
+
+raw_input_list = []
+input_list = []
+initial_trigram_key = tuple()
```
PythonCHB, line 32: tuples are immutable, you never want to "initialize" one like this. It can never be changed once created anyway -- it will get created when it's needed.

students/EricAdams/session04/trigrams.py
```diff
@@ -1,16 +1,101 @@
...
-trigrams = {}
-for i in range(len(words)-2):
-    pair = tuple(words[i:i+2])
-    follower = words[i+2]
-    print(pair, follower)
-    if pair in trigrams:
-        trigrams[pair].append(follower)
+import random
+
+raw_input_list = []
```
PythonCHB, line 30: you don't need to initialize variables in Python. YOu may want to do this if you are going to start with an empty list, and then put stuff in it --but in this case you are reassigning this same name to a brand new list later on anyway, so this is a worthless line.

students/EricAdams/session04/trigrams.py
```diff
@@ -1,16 +1,101 @@
...
+                '"',
+                '?',
+                ';',
+                '\n']
+
+# Put the input file into a list,(by char)
+
+
+with open('./sherlock.txt', 'r') as f_input:
+    raw_input_list = list(f_input.read())
```
PythonCHB, line 50: `.read()` on a text file returns the entire file as one big string. calling list() on that turns it into a list of individual characters, which you probably don't want.

students/EricAdams/session04/trigrams.py
```diff
@@ -1,16 +1,101 @@
...
+
+# Build up an input list with unwanted punctuation replaced
+for i, x in enumerate(raw_input_list):
```
PythonCHB, line 53: if you had kept it in a big string, you could then use the string `.replace()` method to replace characters, which is going to be MUCH faster.

Or, even better, use the string `.translate` method to replace a bunch all at once.

PythonCHB, line 53: you aren't using the index, so no need for `enumerate`

students/EricAdams/session04/trigrams.py
```diff
@@ -1,16 +1,101 @@
...
+# print(input_list)
+
+# Create the trigram from the cleaned up input
+trigram = {}
+# we need a pair hence the len(input_list) - 2 instead of len(input_list) - 1
+for i in range(len(input_list) - 2):
+    # must be a tuple in order to be a key in trigram
+    trigram_key = tuple(input_list[i:i + 2])
+    trigram_value = input_list[i + 2]
+    # this is so nifty, I had to use it
```
PythonCHB, line 71: :-)

students/EricAdams/session04/trigrams.py
```diff
@@ -1,16 +1,101 @@
...
+    trigram.setdefault(trigram_key, []).append(trigram_value)
+
+# Build the text output.text
+# Start with a random word from the trigram
+
+random_number = random.randrange(0, len(trigram) - 1)
```
PythonCHB, line 77: cool -- I've never seen randrange before. But it's docs say:

Choose a random item from ``range(start, stop[, step])``.

This fixes the problem with ``randint()`` which includes the
endpoint; in Python this is usually not what you want.

Which means you don't need the -1 !

students/EricAdams/session04/trigrams.py
```diff
@@ -1,16 +1,101 @@
...
+for i, j in enumerate(trigram):
```
PythonCHB, line 78: This works, but is not a very efficient way to get a random key -- you have to loop though much (on average half) of the entire dict.

Better to use the nifty random.choice() on the keys:

`initial_trigram_key = random.choice(list(trigram.keys()))`

you need the `list()` call, as a dict keys object isn't indexable, so can't be passed in to `random.choice()`

students/EricAdams/session04/trigrams.py
```diff
@@ -1,16 +1,101 @@
...
+    if i == random_number:
+        initial_trigram_key = j
```
PythonCHB, line 80: probably want to use a more meaningful name than "j" -- j is usually used for simple integers like an index.

students/EricAdams/session04/trigrams.py
```diff
@@ -1,16 +1,101 @@
...
+    # Next keyword pair = last two elements of output_list
+    new_key = tuple(output_list[-2:])
+    try:
+        trigram_value_list = trigram[new_key]
+    except KeyError:
+        # pick a random word in trigram_value_list
+        print('exiting')
+        break
+    random_trigram_value = random.choice(trigram_value_list)
+    new_value = random_trigram_value
```
PythonCHB, line 94: why do you need two names for this?

students/EricAdams/session04/trigrams.py
```diff
@@ -1,16 +1,101 @@
...
+    output_list = output_list + list(new_key) + [new_value]
```
PythonCHB, line 95: Bingo!

new_key is the last two words in the output_list -- so you don't want to add them to it again!

PythonCHB (commented):
You've almost got it -- certainly the key logic.

I would think about using names that are about what the variable IS, rather than how it is used. i.e.:

"followers" rather than "value"

I also use plural names for lists of things, and singular for single things.

PythonCHB:
I'm going to merge this so we'll know what's new the next time you submit a PR.

PR #105, opened 2017-11-05
Trigram code added

chops77:


students/Chao/session04/trigram.py
```diff
@@ -1,5 +1,71 @@
...
+# An empty list for word pool
+wpool = []
+
+def build_pool(inf):
+    """ Build a pool of words from the input text file """
+    with open(inf, 'r') as lines:
+        # strip the line, make it all lower cases, remove punctuations, then make a new list
+        for line in lines:
+            wpool.extend(line.strip().lower().translate(str.maketrans('', '', string.punctuation)).split())
+
```
codedragon, line 17: Nice one liner. :) Note that because you are putting the entire file into your list, you will end up with stuff that isn't part of the actual story (headers, dates, intro stuff).

students/Chao/session04/trigram.py
```diff
@@ -1,5 +1,71 @@
...
+        # change individual 'i' with 'I'
+        for wordi in range(len(wpool)):
+            if wpool[wordi] == 'i':
+                wpool[wordi] = 'I'
+
```
codedragon, line 22: Note that I is not the only word that may be capitalized in the middle of a sentence. It may make more sense to be more careful about what words you turn to lower case, rather than try to add back in capitalization for proper nouns. Note also that you are going back over the entire list to do this, which could get costly with a large text file.

students/Chao/session04/trigram.py
```diff
@@ -1,5 +1,71 @@
...
+        # Set flag for capitalization
+        if pun == ',':
+            capflag = False
+        else:
+            capflag = True
+        news = news + ' ' + ' '.join(newl)
+    # Make sure the article ends with a period
+    news = news[:-1] + '.'
+    return news
+
```
codedragon, line 63: Nice job. It works!

students/Chao/session04/trigram.py
```diff
@@ -1,5 +1,71 @@
 #!/usr/bin/env python
 
+import string
+import random
+
 # Create an empty dictionary
 trig = {}
+# An empty list for word pool
+wpool = []
+
```
codedragon, line 10: These variables don't need to be here. It would be better if they were initiated in the functions that use them, and then passed as input/output variables. Really, the only variables that should be outside of the functions are constants, iow variables that always have the same value.

codedragon (approved):
Great job.

PR #106, opened 2017-11-06
trigram.py.  

ea8606:
Seems to be stuck in a continuous loop so added a break for 2500 chars


students/EricAdams/session04/trigrams.py
```diff
@@ -1,101 +1,166 @@
...
+    # for i, x in enumerate(raw_input_list):
+    #     if x in replace_punc:
+    #         input_list.append(' ')
+    #     else:
+    #         input_list.append(x)
+
+    # maketrans returns a table
+    transition_table = raw_input_string.maketrans(
+        unwanted_punc, substitution_chars)
+    input_string = raw_input_string.translate(transition_table)
```
PythonCHB, line 83: nice! -- .translate is pretty nifty, isn't it?

students/EricAdams/session04/trigrams.py
```diff
@@ -1,101 +1,166 @@
...
+    :return = a dictionary, (trigram)
+    """
+    trigram = {}
+    # we need a pair hence the len(input_list) - 2
+    #   instead of len(input_list) - 1
+    for i in range(len(input_string) - 2):
+        # must be a tuple in order to be a key in trigram
+        trigram_key = tuple(input_string[i:i + 2])
+        trigram_followers = input_string[i + 2]
+        # this is so nifty, I had to use it
```
PythonCHB, line 99: :-)

students/EricAdams/session04/trigrams.py
```diff
@@ -1,101 +1,166 @@
...
+def initial_output_list(trigram):
+    """Initialize the output list with three elements derived from a
+    random trigram key
+    :param = dictionary, (trigram)
+    :return list
+    """
+    # Remove - 1
+    # random_number = random.randrange(0, len(trigram) - 1)
+    random_number = random.randrange(0, len(trigram))
+    for i, trigram_key in enumerate(trigram):
```
PythonCHB, line 113: kinda klunky, but it works :-)

As we talked about in class, I like:

initial_trigram_key = random.choice(list(trigram))

but that does require making a list of all the keys ahead of time -- could be memory intensive.

students/EricAdams/session04/trigrams.py
```diff
@@ -1,101 +1,166 @@
...
+        if i == random_number:
+            initial_trigram_key = trigram_key
+            break
+    output_list = list(initial_trigram_key) + trigram[initial_trigram_key]
```
PythonCHB, line 117: so here you are starting with your random pair of words -- perfect.

But then you're adding the list of following words -- which you don't want. You want only one of those words.

And now that you have an initial pair, you probably just want to use the regular loop to add the rest...

students/EricAdams/session04/trigrams.py
```diff
@@ -1,101 +1,166 @@
...
+        # raw_input_list = list(f_input.read())
+        raw_input_string = f_input.read()
+
+    # Clean up the input
+    input_string = replace_unwanted_punctuation(
+        raw_input_string, unwanted_punc, substitution_chars)
+
+    # get a list of words, instead of chars
+    input_text = "".join(input_string)
+    input_string = input_text.split(" ")
```
PythonCHB, line 134: If you call `split()` with no argument, it splits on all whitespace.

students/EricAdams/session04/trigrams.py
```diff
@@ -1,101 +1,166 @@
...
-t = " ".join(output_list)
-# print(t)
-with open('trigrams.txt', 'w') as f_output:
-    f_output.write(t)
+    while(True):
+        # Next keyword pair = last two elements of output_list
+        new_key = tuple(output_list[-2:])
+        try:
+            trigram_followers_list = trigram[new_key]
+        except KeyError:
```
PythonCHB, line 149: of course, as you noticed, this  might never happen!

students/EricAdams/session04/trigrams.py
```diff
@@ -1,101 +1,166 @@
...
+            print('exiting')
+            break
+        random_trigram_followers = random.choice(trigram_followers_list)
```
PythonCHB, line 152: careful with your variable names -- this is a single word, so confusing to use a plural name.

I would probably call the list simply "followers" - the plural indicates it's more than one -- and then when you pull out a random one, maybe "follower" or "next_word" or something singular.

students/EricAdams/session04/trigrams.py
```diff
@@ -1,101 +1,166 @@
...
+        # useless line
+        # new_value = random_trigram_followers
+        # Don't add list(new_key) to output
+        # output_list = output_list + list(new_key) +
+        # [random_trigram_followers]
+        # output_list = output_list + list(new_key[1]) +
+        # [random_trigram_followers]
+        output_list = output_list + [random_trigram_followers]
```
PythonCHB, line 160: if you are adding one element to a list -- append() is the way to go. This works fine, but it's creating a whole new list just to throw it away again:

output_list.append (random_trigram_followers)

and if you are adding another list of objects, you want to use extend(). The way it is now, you are actually doing:

1) create a list with random_trigram_followers in it.
2) create a new list with the contents of output_list the new list.
3) re-assign the name output_list to that new list.
4) throw away the original output_list and the temporary list you created for random_trigram_followers

That's a LOT of overhead!

Be cognizant of when you are mutating (changing) an object vs creating a new one and assigning it to the same name. When you do assignment (=) you are probably creating a new object.

+= is different -- it is the "in_place" operator, so:

a_list += another_list

does not create an new lists -- it adds to the original list "in place" -- it is identical to:

a_list.extend(another_list)

And is an efficient operation.

students/EricAdams/session04/trigrams.py
```diff
@@ -1,101 +1,166 @@
...
+        if len(output_list) > 5000:
```
PythonCHB, line 161: yup -- you need a termination critera!

PythonCHB (approved):
a few things to tweak, but it looks good.

Nice work

PR #107, opened 2017-11-06
mailroom for review

kmarguerite:
Mailroom now has a few comprehensions and exception handling, but I need some review. 

I'm running into a problem printing my report after adding a new person via the thank you letter. It throws an error indicating float is not iterable, the float is coming from a dictionary. Searching on stack overflow, recommended to append new keys/values to the list. I tried this but couldn't get it to work successfully, still working on this issue. Any tips?

Thank you for your help!

Kate


PythonCHB:
Sorry it took so long to get to this!

In the future, if you have a specific question, and we don't respond quickly via teh PR, feel free to ping us via email or Slack.

students/kate/session05/mailroom3.py
```diff
@@ -0,0 +1,136 @@
...
+
+      > ''')
+    return action.strip()
+
+def write_thankyou(donor_input, amount):
+    return('''
+          Dear {}
+
+          Thank you for your very donation of ${:.2f}.
+          You have enabled excellemt magical instruction!
```
PythonCHB, line 34: typo in "excellemt" :-)

students/kate/session05/mailroom3.py
```diff
@@ -0,0 +1,136 @@
...
+        if amount == "quit":
+            return
+        try:
+            amount = float(amount)
+        except ValueError:
+            print("Invalid entry! Please enter a number!")
+        else:
+            break
+
+    donors_dict[donor_input] = amount
```
PythonCHB, line 72: Here's you issue ( I think :-) ):

amount is a float -- perfect type for, well, an amount of something :-)
but donors_dict is a dict with the names as the keys, and a LIST of amounts as the value.

With this line, you are replacing a list of values with a single value -- which is going to cause problems later when you need to work with that list (plus you will have thrown away any previous donations)

You want: 

`donors_dict[donor_input].append(amount)`

if this is a new donor, you want to have put an empty list in there before this line.

students/kate/session05/mailroom3.py
```diff
@@ -0,0 +1,136 @@
...
+        # do a case-insenstive compare
+        # if name.strip().lower() == donor[0].lower():
+#            return donor
+#    return None
+
+
+# loop through the donor list and print the 0th element of the list
+def print_donors_list():
+    print("Donors:\n")
+    for key in donors_dict:
```
PythonCHB, line 89: I like to use names that are meaningful to this data, rather than generic to dictionaries:

maybe:

```
for name in donors_dict:
    print(name)
```

students/kate/session05/mailroom3.py
```diff
@@ -0,0 +1,136 @@
...
+        print(key)
+
+# this forces the first item, not the 0th item, to be the first item
```
PythonCHB, line 92: well, not the first item, but the item that gets used for sorting. We call that the "sort key".

PythonCHB (approved):
One tiny error -- the rest looks great.

PR #108, opened 2017-11-06
johnn session05 labs

johnnzz:
* Reflow session04 assignments after fixing tab settings in editors.
* Add session05 labs.
* Add a bit more exception handling to mailroom.
* Add some unit tests to mailroom.

It is a known deficiency that the functions need to be refactored to support objects to be passed in to facilitate testing.

### DIFF
--- a/students/john_navitsky/session04/dict_lab.py
+++ b/students/john_navitsky/session04/dict_lab.py
@@ -24,8 +24,8 @@ print( "Mango" in person.values() )
 # create a new dict from old
 new_dict = {}
 for key in person:
-	old_value=person[key].lower()
-	new_dict[key]=old_value.count("t")
+    old_value=person[key].lower()
+    new_dict[key]=old_value.count("t")
 print(new_dict)
 
 # for convience create a dict for the sets
@@ -33,13 +33,13 @@ s = { 2: set(), 3: set(), 4: set() }
 
 # loop through the combinations and build up the sets
 for item in range(21):
-	for entry in range(2,5):
-		if item % entry == 0:
-			s[entry].add(item)
+    for entry in range(2,5):
+        if item % entry == 0:
+            s[entry].add(item)
 
 # print sets
 for item in range(2,5):
-	print(item,s[item])
+    print(item,s[item])
 
 # some tests
 print("s3 is a subset of s2:", s[2].issubset(s[3]))
--- a/students/john_navitsky/session04/mailroom.py
+++ b/students/john_navitsky/session04/mailroom.py
@@ -212,14 +212,14 @@ def thank_you_entry():
 
 
 def thank_all_donors():
-	for index in range(len(donors)):
-		#dest=sys.stdout
-		donor_name = donors[index]["last_name"]+"_"+donors[index]["first_name"]
-		donor_name = donor_name.strip().lower().replace(" ", "_").replace(",", "")
-		donor_file="thank_you_" + donor_name
-		dest = open(donor_file, "w")
-		print_thank_you(index,"wonderful",dest)
-		dest.close()
+    for index in range(len(donors)):
+        #dest=sys.stdout
+        donor_name = donors[index]["last_name"]+"_"+donors[index]["first_name"]
+        donor_name = donor_name.strip().lower().replace(" ", "_").replace(",", "")
+        donor_file="thank_you_" + donor_name
+        dest = open(donor_file, "w")
+        print_thank_you(index,"wonderful",dest)
+        dest.close()
 
 def main():
     """ Main menu / input loop. """
@@ -249,7 +249,7 @@ def main():
             print_report()
 
         if selection in ["p", "print"]:
-        	thank_all_donors()
+            thank_all_donors()
 
         # accept either send or enter
         if selection in ["2", "s", "send", "e", "enter"]:
--- a/students/john_navitsky/session04/trigrams.py
+++ b/students/john_navitsky/session04/trigrams.py
@@ -35,8 +35,8 @@ word_list = text.split()
 # lower case words following periods because they 
 # are probably not proper nouns
 for index, word in enumerate(word_list):
-	if word == "." and index < len(word_list)-1:
-		word_list[index+1]=word_list[index+1].lower()
+    if word == "." and index < len(word_list)-1:
+        word_list[index+1]=word_list[index+1].lower()
 
 #print()
 #print("PROCESSED_LIST:",word_list)
@@ -44,11 +44,11 @@ for index, word in enumerate(word_list):
 # index pairs
 pairs={}
 for item in range(2,len(word_list)):
-	pair=( word_list[item-2],word_list[item-1] )
-	follower=word_list[item]
-	if pair not in pairs.keys():
-		pairs[pair]=[]
-	pairs[pair].append(follower)
+    pair=( word_list[item-2],word_list[item-1] )
+    follower=word_list[item]
+    if pair not in pairs.keys():
+        pairs[pair]=[]
+    pairs[pair].append(follower)
 
 #print()
 #print("PAIRS:",pairs)
@@ -61,26 +61,26 @@ out_list=list(random.choice(list(pairs.keys())))
 #print()
 #print("writing the story",end="")
 while True:
-	cur_key=( out_list[i-1], out_list[i] )
-	if cur_key in pairs.keys():
-		next_word=random.choice(pairs[cur_key])
-		out_list.append(next_word)
-		i+=1
-		#print(".",end="")
-	else:
-		break
-	# don't let the output get too big
-	if i > len(word_list) and next_word == ".":
-		# break on period for continuity
-		break
+    cur_key=( out_list[i-1], out_list[i] )
+    if cur_key in pairs.keys():
+        next_word=random.choice(pairs[cur_key])
+        out_list.append(next_word)
+        i+=1
+        #print(".",end="")
+    else:
+        break
+    # don't let the output get too big
+    if i > len(word_list) and next_word == ".":
+        # break on period for continuity
+        break
 
 #print("GENERATED:",out_list)
 
 # first word of sentence get capped
 for index, word in enumerate(out_list):
-	# careful not to walk off the end
-	if word in [".", "?", "!"] and index < len(out_list)-1:
-		out_list[index+1]=out_list[index+1].capitalize()
+    # careful not to walk off the end
+    if word in [".", "?", "!"] and index < len(out_list)-1:
+        out_list[index+1]=out_list[index+1].capitalize()
 
 # capitalize the very first word
 out_list[0]=out_list[0].capitalize()
--- a/students/john_navitsky/session05/except_exercise.py
+++ b/students/john_navitsky/session05/except_exercise.py
@@ -1,0 +1,60 @@
+#!/usr/bin/python
+
+"""
+An exercise in playing with Exceptions.
+Make lots of try/except blocks for fun and profit.
+
+Make sure to catch specifically the error you find, rather than all errors.
+"""
+
+from except_test import fun, more_fun, last_fun
+
+
+# Figure out what the exception is, catch it and while still
+# in that catch block, try again with the second item in the list
+first_try = ['spam', 'cheese', 'mr death']
+
+try:
+    joke = fun(first_try[0])
+    joke = fun(first_try[1])
+except NameError:
+    print("Whoops!  There is no joke for: spam")
+
+# Here is a try/except block. Add an else that prints not_joke
+try:
+    not_joke = fun(first_try[2])
+except SyntaxError:
+    print('Run Away!')
+else:
+    print(not_joke)
+
+# What did that do? You can think of else in this context, as well as in
+# loops as meaning: "else if nothing went wrong"
+# (no breaks in  loops, no exceptions in try blocks)
+
+# Figure out what the exception is, catch it and in that same block
+#
+# try calling the more_fun function with the 2nd language in the list,
+# again assigning it to next_joke.
+
+# If there are no exceptions, call the more_fun function with the last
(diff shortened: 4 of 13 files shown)


students/john_navitsky/session04/dict_lab.py
```diff
@@ -24,22 +24,22 @@
 # create a new dict from old
 new_dict = {}
 for key in person:
-	old_value=person[key].lower()
-	new_dict[key]=old_value.count("t")
+    old_value=person[key].lower()
+    new_dict[key]=old_value.count("t")
```
PythonCHB, line 28: this is the kind of thing it makes sense to make a one-liner (andif you need the keys and values, loop through them both at once:

```
for key, value in person.items():
    new_dict[key] = value.lower().count('t') 
```

students/john_navitsky/session04/mailroom.py
```diff
@@ -212,14 +212,14 @@ def thank_you_entry():
...
-		donor_name = donors[index]["last_name"]+"_"+donors[index]["first_name"]
-		donor_name = donor_name.strip().lower().replace(" ", "_").replace(",", "")
-		donor_file="thank_you_" + donor_name
-		dest = open(donor_file, "w")
-		print_thank_you(index,"wonderful",dest)
-		dest.close()
+    for index in range(len(donors)):
+        #dest=sys.stdout
+        donor_name = donors[index]["last_name"]+"_"+donors[index]["first_name"]
+        donor_name = donor_name.strip().lower().replace(" ", "_").replace(",", "")
```
PythonCHB, line 218: you replace the spaces with underscores (good idea for a filename), but then add a space after the commas.

Why?

PythonCHB:
most of the changes were indentation cleanup -- so if I missed commenting on something you want reviewed, let me know.

students/john_navitsky/session04/trigrams.py
```diff
@@ -35,20 +35,20 @@
...
 
 # index pairs
 pairs={}
 for item in range(2,len(word_list)):
-	pair=( word_list[item-2],word_list[item-1] )
-	follower=word_list[item]
-	if pair not in pairs.keys():
-		pairs[pair]=[]
-	pairs[pair].append(follower)
+    pair=( word_list[item-2],word_list[item-1] )
```
PythonCHB, line 47: you can do this a bit cleaner with a slice:

`pair= tuple(word_list[item - 2: item])`

students/john_navitsky/session04/trigrams.py
```diff
@@ -35,20 +35,20 @@
...
+    follower=word_list[item]
+    if pair not in pairs.keys():
```
PythonCHB, line 49: this can jsut be:

if pair not in pairs:

"in" (and not in) for adict means "is it a key in the dict?"

students/john_navitsky/session04/trigrams.py
```diff
@@ -61,26 +61,26 @@
...
-		break
-	# don't let the output get too big
-	if i > len(word_list) and next_word == ".":
-		# break on period for continuity
-		break
+    cur_key=( out_list[i-1], out_list[i] )
+    if cur_key in pairs.keys():
+        next_word=random.choice(pairs[cur_key])
+        out_list.append(next_word)
+        i+=1
```
PythonCHB, line 68: as you are building this as you go -- you don't really need the index.

You can access the last two words in the list with slicing:

`cur_key = tuple(out_list[-2:])`

-CHB

students/john_navitsky/session05/except_exercise.py
```diff
@@ -0,0 +1,60 @@
...
+
+# Figure out what the exception is, catch it and while still
+# in that catch block, try again with the second item in the list
+first_try = ['spam', 'cheese', 'mr death']
+
+try:
+    joke = fun(first_try[0])
+    joke = fun(first_try[1])
+except NameError:
+    print("Whoops!  There is no joke for: spam")
```
PythonCHB, line 21: can you call it with first_try[2] now?

students/john_navitsky/session05/except_exercise.py
```diff
@@ -0,0 +1,60 @@
...
+
+# call more_fun() with each lang
+for lang in langs:
+    try:
+        next_joke = more_fun(lang)
+    except IndexError:
+        continue
+    finally:
+        # it doesn't appear from the assignment sample output it was
+        # envisioned this would be in a loop, so only run last_fun()
```
PythonCHB, line 57: I actually hard coded the langs[0] and langs[1], rather than a loop...

but if you use a loop, then you'll want to put the loop inside the try block. But the goal was to get you to write a try block with every feature -- try, except, else, and finally....

(hard to come up with a simple example that uses all that!)

```
langs = ['java', 'c', 'python']
try:
    more_joke = more_fun(langs[0])
except IndexError:
    more_joke = more_fun(langs[1])
else:
    more_joke = more_fun(langs[-1])
finally:
    last_fun()
```

students/john_navitsky/session05/mailroom.py
```diff
@@ -0,0 +1,328 @@
...
+
+def safe_input(prompt=">"):
+    """ Generic input routine. """
+    # return null if anything goes wrong
+    selection=""
+    try:
+        selection=input(prompt).strip()
+    except (KeyboardInterrupt, EOFError):
+        # don't exit the program on ctrl-c, ctrl-d
+        pass
```
PythonCHB, line 39: This is going to go to "return selection" if one of those Exceptions is raised -- but selection won't be defined.

students/john_navitsky/session05/mailroom.py
```diff
@@ -0,0 +1,328 @@
...
+
+    print_lines()
+  
+    # print report header
+    #      index    name   total gvn #gifts  avg gift
+    print("{0:5} | {1:20} | {2:14} | {3:9} | {4:12}".format(
+        "ID","Donor Name","Total Given","Num Gifts","Average Gift"))
+    print("-"*72)
+
+    # enumerate so we have the list index handy
```
PythonCHB, line 61: Is the index important? you are using an ordered collection (list), but I don't think order has any significance here -- I'd leave the index out of it.

If you think of it as an arbitrary collection, it'll be easier to refactor into another storage scheme.

students/john_navitsky/session05/mailroom.py
```diff
@@ -0,0 +1,328 @@
...
+
+        average_donation = total_donations / num_donations
+
+        #      index      name     total gvn    #gifts    avg gift
+        print("{0:05d}   {1:20}  ${2:14,.2f}   {3:9.0f}  ${4:12,.2f}".format(
+            donor_index,donor["first_name"]+" "+donor["last_name"],
+            total_donations,num_donations,average_donation))
+
+
+def parse_name(full_name):
```
PythonCHB, line 78: nice!

students/john_navitsky/session05/mailroom.py
```diff
@@ -0,0 +1,328 @@
...
+        donor_file="thank_you_" + donor_name
+        dest = open(donor_file, "w")
+        print_thank_you(index,"wonderful",dest)
+        dest.close()
+
+def main():
+    """ Main menu / input loop. """
+    
+
+    menu =  "\n"
```
PythonCHB, line 241: this is a good use for a triple quoted string:

```
menu = """
DONATION WIZARD MAIN MENU\n"
-------------------------\n"

Select from the following:
...
```
Or -- if you do build it up in pieces, put them all in a list and then join() them (you could join with '\n' in this case)

Or:

Python concatenates strings if they are one after another, and continues lines if in a parens, so:

```
menu = ("DONATION WIZARD MAIN MENU\n"
        "-------------------------\n\n"
        "Select from the following:\n\n"
        ....
```

students/john_navitsky/session05/mailroom.py
```diff
@@ -0,0 +1,328 @@
...
+
+    selection=None
+    while selection not in ["0", "quit", "q"]:
+
+        print_lines()
+
+        print(menu)
+        selection=safe_input("(l)ist, (e)nter, (q)uit: ").lower()
+
+        if selection in ["1", "l", "list"]:
```
PythonCHB, line 261: nice!

I tend to use tuples, rather that lists:

`if selection in ("1", "l", "list"):`

since you won't be changing it. and a tuple is a tiny bit lighter weight than a list.

students/john_navitsky/session05/mailroom.py
```diff
@@ -0,0 +1,328 @@
...
+    print_lines()
+
+
+def print_fatal(routine="importaint"):
+    """ Upon fatal error, print message and exit. """
+    print("The {} function has malfunctioned.  Contact customer support!".format(routine))
+    sys.exit(1)
+
+
+def sanity_tests():
```
PythonCHB, line 286: nice -- you're going to move this to your test code now, yes?

but you totally got the idea.

students/john_navitsky/session06/mailroom.py
```diff
@@ -0,0 +1,319 @@
+#!/usr/bin/env python
+
+import sys
+import json
+
+""" Program to manage donations. """
+
+def load_donor_file(donor_file="donors.json"):
```
PythonCHB, line 8: very nicely done!

students/john_navitsky/session06/mailroom.py
```diff
@@ -0,0 +1,319 @@
...
+        with open(donor_file,'w') as donor_data:
+            donor_data.write(json.dumps(donors))
+        return True
+    except (PermissionError,OSError) as e:
+        print("Sorry, couldn't write the donor file!")
+        print(e)
+        return False
+
+
+def safe_input(prompt=">",mock=False,mock_in=""):
```
PythonCHB, line 47: you really don't want to have mocking be part of your actual code....

It's a tricky business, but the mock module can inject stuff in for you in tests.

johnnzz:
@PythonCHB the only thing of interest here is some session06 stuff that got piggy-backed here:

https://github.com/UWPCE-PythonCert/IntroPython-2017/pull/108/commits/6185a2b2670569d8c74b5483d27a11746978d21b

students/john_navitsky/session06/mailroom.py
```diff
@@ -0,0 +1,319 @@
...
+        except (KeyboardInterrupt, EOFError):
+            # don't exit the program on ctrl-c, ctrl-d
+            selection=""
+        return selection
+    else:
+        # return what the test program told you to
+        return mock_in
+
+
+def print_lines(lines=2,dest=sys.stdout):
```
PythonCHB, line 63: if you are going to make the output destination settable (which is a great idea) then you may jsut want to go with file writing methods, rather than print() (sys.stdout is a "file-like" object.

```
for i in range(lines):
    dest.write("\n")
```
and when you look at it that way:

`dest.write(lines * "\n")`

:-)

students/john_navitsky/session06/test_mailroom.py
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env python3
+
+import mailroom
+import pytest
+import io
+
+
+def test_true():
```
PythonCHB, line 8: :-)

students/john_navitsky/session06/test_mailroom.py
```diff
@@ -0,0 +1,43 @@
...
+    assert True
+
+
+def test_parse_name():
+    assert mailroom.parse_name("Mary Jo Smith, IV") == {
```
PythonCHB, line 13: just a note, if you're curious, for a "Full" name structure:

First, Middle, von, Last, Jr

at least this is how BiBTeX (bibliography management software) handles it --it's pretty universal actually, at least for European names.

students/john_navitsky/session06/test_mailroom.py
```diff
@@ -0,0 +1,43 @@
...
+        'full_name': 'Mary Jo Smith, Iv', 
+        'informal_name': 'Mary Jo Smith',
+        'suffix': 'IV',
+        'last_name': 'Smith',
+        'first_name': 'Mary Jo' }
+
+
+def test_print_lines():
+    # verify the print_lines() function
+    out = io.StringIO()
```
PythonCHB, line 23: nice find! -- exactly the right tool for testing this kind of thing.

students/john_navitsky/session06/test_mailroom.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+# def test_print_thank_you():
+#      # verify the print_thank_you() function
+#     out = io.StringIO()
+#     mailroom.print_thank_you(0,"testfull",out)
+#     output = out.getvalue()
+#     out.close()
+#     assert "Dearest Joe Smith," in output
+
```
PythonCHB, line 42: good start on the testing.

PythonCHB (approved):
looking great!

students/john_navitsky/session04/mailroom.py
johnnzz: Replace commas with nulls.  Normally, I wouldn't want to store the name as a single string, containing commas, ("Mary Jo Smith, III"), but I felt like it was outside of the parameters of the exercise to change that.

So names can have commas in this model, and when writing the file, you could have "Mary Joe Smith" and "Mary Joe Smith, III", so you don't just want to ditch the suffix.

So I just glue it all together without commas*:

```
In [4]: parse_name("Mary Jo Smith, III")
Out[4]:
{'first_name': 'Mary Jo',
 'full_name': 'Mary Jo Smith, Iii',
 'informal_name': 'Mary Jo Smith',
 'last_name': 'Smith',
 'suffix': 'III'}

In [5]:
```
And you get:
```
johnn@johnn-lnx:~/uwpython/IntroPython-2017/students/john_navitsky/session06$ ls thank*
thank_you_smith_iii_mary_jo
johnn@johnn-lnx:~/uwpython/IntroPython-2017/students/john_navitsky/session06$
```

* Yes, I see that caps bug.

PR #109, opened 2017-11-07
session05 files and updates to mailroom for comprehensions

Ginsusamurai:
updated doc strings and added mailroom input file


students/morgan/session05/except_exercise.py
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/python
```
PythonCHB, line 1: you got it.

PythonCHB (approved):
If I missed something you want me to review, let me know.

PR #110, opened 2017-11-07
adding work so far for Session 05

mattbriggs:


students/briggsm/session04/mb_trigram.py
```diff
@@ -0,0 +1,54 @@
+import random as rand
+
+def create_corpus(infile):
```
PythonCHB, line 3: love the word "corpus" !

students/briggsm/session04/mb_trigram.py
```diff
@@ -0,0 +1,54 @@
...
+
+def create_trigram(textlist):
+    '''Input: a list of words from a text body.
+       Output: a dictionary of two words in sequence and the word after.'''
+    trigram = {}
+    for w in range(len(textlist[:-2])):
+        firstword = textlist[w]
+        secondword = textlist[w+1]
+        thirdword = textlist[w+2]
+        trigram[thirdword] = (firstword, secondword)
```
PythonCHB, line 24: I think this is backwards -- you have the third word as the key and the first two as the value, so now you can look up a word and see what two words preceded it. Which is not what you want.

Plus, every time you hit the same word again, it will overwrite the last occurrence!

you (almost) want, in the end, something like:

trigram[(firstword, secondword)] = [a_previous_word, another_one, third_word]

that is -- the keys are the first two words, and the value is a list of all the words that have followed that pair of words.

students/briggsm/session04/mb_trigram.py
```diff
@@ -0,0 +1,54 @@
...
+    return trigram
+
+def get_values(trigram):
+    '''Input: a dictionary.
+       Output: a list of keys.'''
+    words = []
+    for w in trigram.keys():
```
PythonCHB, line 31: this is the same as:

`words = list(trigram)`

but it's not what you want -- it makes a list of all the keys in the dict -- not really what you want. with the dict you built, I think this is going to give you an arbitrarily ordered list of all the works in the original text.

students/briggsm/session04/mb_trigram.py
```diff
@@ -0,0 +1,54 @@
...
+        words.append(w)
+    return words
+
+def get_random_item(listofwords):
+    '''Input: a list of items.
+       Output: a random item as a string.'''
+    choose = rand.randint(0,len(listofwords))
+    return str(listofwords[choose])
```
PythonCHB, line 39: this works, but there's the:

`random.choice(listofwords)`

function which does the same thing.

students/briggsm/session04/mb_trigram.py
```diff
@@ -0,0 +1,54 @@
...
+
+def create_random_text(listofwords, trigram, wordcount):
```
PythonCHB, line 41: re-factor how you build the dict and then re-visit this.

the way it is, you are getting random text that uses the same words, and some of the same work pairings, but note getting what we are looking for:

each word pair is followed by one of the words that that same pair was followed by in the original text.

students/briggsm/session05/mailroom3.py
```diff
@@ -0,0 +1,109 @@
+'''Session 5 Mailroom3
```
PythonCHB, line 1: I think I mostly reviewed this in its earlier version.

let me know if there's something new you want me to look at.

PythonCHB (commented):
trigrams logic is not quite right -- but I think you can fix it while still re-using a fair bit of your code.

mailroom looks great!

PythonCHB:
I'm merging so it'll be easy to see what's new the next time.

-CHB

PR #112, opened 2017-11-09
mailroom with tests; kwargs exercise; html renderer project

mattmaeda:


students/mattmaeda/.gitignore
```diff
@@ -0,0 +1,2 @@
+*.pyc
```
PythonCHB, line 1: is that not in the master .gitignore?

not that it hurts to duplicate it.

students/mattmaeda/session06/mailroom.py
```diff
@@ -0,0 +1,120 @@
+#!/usr/bin/env python3
+"""
+Mailroom project for Intro to Python Class
+"""
+
+REPORT_HEADER = "{donor: <50}| {total: <12}| {num: <10}| {avg: <12}"
+REPORT_LINE = "{0: <50} ${1: 12.2f}{2: 12d}{3: 14.2f}"
+
+class Mailroom(object):
```
PythonCHB, line 9: ahh -- jumping ahead here :-)

we'll be talking about making mailroom OO soon enough. I'll review it then.

And you don't need to inherit from object unless you need python2 compatibility.

students/mattmaeda/session06/test_mailroom.py
```diff
@@ -0,0 +1,60 @@
...
+            "John Quincy Adams": [2],
+            "James Madison": [1]
+        }
+
+        self.mailroom = Mailroom(donors)
+
+
+    def test_get_menu(self):
+        """ test menu """
+        self.assertEqual({1: "Send a Thank You",
```
PythonCHB, line 33: you want to be careful with tests like this -- it's expecting EXACTLY that result -- so if the code is extended to add another menu item, It'll break, even though the functionality of the method hasn't broken.

This is  a good chance to think about whether there is any point in a get_menu method at all.

students/mattmaeda/session06/test_mailroom.py
```diff
@@ -0,0 +1,60 @@
...
+Monty Python                                       $       10.00           1         10.00
+John Adams                                         $        3.00           1          3.00
+Thomas Jefferson                                   $        3.00           1          3.00
+John Quincy Adams                                  $        2.00           1          2.00
+George Washington                                  $        1.00           1          1.00
+James Madison                                      $        1.00           1          1.00
+
+"""
+
+class TestMailroom(unittest.TestCase):
```
PythonCHB, line 17: why unittest?

you might want to try pure pytest tests a bit --much less verbose :-)

students/mattmaeda/session06/test_mailroom.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+    def test_thanks(self):
+        """ test thank you letter """
+        self.assertEqual("\n\nDear Monty Python,\nThank you for your " \
+                         "donation of 10.\n\n",
+                         self.mailroom.thanks("Monty Python", 10))
+
+
+    def test_report(self):
```
PythonCHB, line 45: another case of testing precice output -- rather, try to test that report has qualites it should have:

the right number of rows, the right header, at _some_ of the donors in there. Ideally, of you use a different test DB, the tests shouldn't fail.

Though you did specify the test DB in the tests -- very good practice!

students/mattmaeda/session07/html_render.py
```diff
@@ -0,0 +1,173 @@
+#!/usr/bin/env python3
+# coding=utf-8
```
PythonCHB, line 2: utf-8 is implied with python3

students/mattmaeda/session07/html_render.py
```diff
@@ -0,0 +1,173 @@
...
+from __future__ import unicode_literals
```
PythonCHB, line 3: and this is only for py2 code -- for py2/3 compatibility.

Are you practicing wrting py2/3 compatible code? If so --carry on :-)

students/mattmaeda/session07/html_render.py
```diff
@@ -0,0 +1,173 @@
...
+
+class Element(object):
+    """ class for rendering HTML element
+    """
+    def __init__(self, content=None, **kwargs):
+        self.tag = "tag"
```
PythonCHB, line 9: you want the tag to be a class attribute rather than an instance attribute --that way all Element instances of the same type share the same tag.

students/mattmaeda/session07/html_render.py
```diff
@@ -0,0 +1,173 @@
...
+        self.elements = []
+        self.kwargs = {}
```
PythonCHB, line 11: I wouldn't call the instance attribute "kwargs" -- kwarg is a very python-y thing that mean, well "the kwy word arguments" -- but at this point, they are no longer keyword arguments, but rather element attributes, so probably better to use a name that reflects that.

students/mattmaeda/session07/html_render.py
```diff
@@ -0,0 +1,173 @@
...
+
+        if content is not None:
+            self.append(content)
+
+        if kwargs is not None:
```
PythonCHB, line 16: kwargs will never be None -- if there are no keyword arguments, then it'll be an empty dict -- which is what you want.

students/mattmaeda/session07/html_render.py
PythonCHB, line 11: no need to initialize this -- set it when you set it with the actual kwargs dict.

students/mattmaeda/session07/html_render.py
```diff
@@ -0,0 +1,173 @@
...
+
+    def append(self, content):
+        """ Appends content to class attribute elements list
+        """
+        self.elements.append(content)
+
+
+    def render(self, file_out, ind="", depth=1):
+        """ Writes content to StringIO buffer with optional indentation
+        :file_out StringIO: StringIO
```
PythonCHB, line 27: NOTE: file_out is any "file like" object. it MIGHT be a StringIO buffer, but it might jsut as well be a real file.

Which is the entire point of StringIO :-)

students/mattmaeda/session07/html_render.py
```diff
@@ -0,0 +1,173 @@
...
+        :ind String: char indentation
```
PythonCHB, line 28: the usual way to document parameters and types is:

```
:param file_out: file to write the html to.
:type: file-like object

:param ind: current amount of indentation
:type ind: str
```

students/mattmaeda/session07/html_render.py
```diff
@@ -0,0 +1,173 @@
...
+        if kwargs is not None:
+            self.kwargs = kwargs
+
+    def append(self, content):
+        """ Appends content to class attribute elements list
+        """
+        self.elements.append(content)
+
+
+    def render(self, file_out, ind="", depth=1):
```
PythonCHB, line 25: if you are passing in depth, then no need to pass in indentation also....

the size of indentation should be a class attribute, and the depth is how many to use.

students/mattmaeda/session07/html_render.py
```diff
@@ -0,0 +1,173 @@
...
+        """ Writes content to StringIO buffer with optional indentation
+        :file_out StringIO: StringIO
+        :ind String: char indentation
+        """
+        formatted_open_tag = self.tag
```
PythonCHB, line 30: see if you can build all the pieces up in a list, and the .join() them all -- a little better performing than += for strings.

students/mattmaeda/session07/html_render.py
```diff
@@ -0,0 +1,173 @@
...
+        for key, value in self.kwargs.items():
+            formatted_open_tag += ' {}="{}"'.format(key, value)
+
+        file_out.write("{}<{}>\n".format(ind*depth, formatted_open_tag))
+
+        for elem in self.elements:
```
PythonCHB, line 36: nice and clean.

students/mattmaeda/session07/html_render.py
PythonCHB: If this is py2/3 compatible code -- put a note in the comments at the top of the file.

students/mattmaeda/session07/html_render.py
```diff
@@ -0,0 +1,173 @@
...
+            file_out.write("{}</{}>\n".format(ind*depth, self.tag))
+
+
+class OneLineTag(Element):
+    def __init__(self, content=None, **kwargs):
+        super(OneLineTag, self).__init__(content, **kwargs)
+
+
+    def render(self, file_out, ind = "", depth=1):
+        for elem in self.elements:
```
PythonCHB, line 52: this is assuming there will only be text in this element -- probably be a safe assumption of a one-line tag -- but maybe actually enforce that there is only one piece of content if you're going to do that.

students/mattmaeda/session07/html_render.py
```diff
@@ -0,0 +1,173 @@
...
+            file_out.write("{0}<{1}>{2}</{1}>\n".format(ind*depth, self.tag,
+                                                        elem))
+
+
+class SelfClosingTag(Element):
+    def __init__(self, content=None, **kwargs):
```
PythonCHB, line 58: again -- non need for an __init__ that only calls the superclass __init__.

students/mattmaeda/session07/html_render.py
```diff
@@ -0,0 +1,173 @@
...
+
+
+    def render(self, file_out, ind = "", depth=1):
+        file_out.write("{}<{} />\n".format(ind*depth, self.tag))
+
+
+class Html(Element):
+    """ Subclass of element for HTML tags
+    """
+    def __init__(self, content=None, **kwargs):
```
PythonCHB, line 69: if you make the tag a class attribute, then no need for an __init__ at all.

students/mattmaeda/session07/html_render.py
```diff
@@ -0,0 +1,173 @@
...
+        super(Html, self).__init__(content, **kwargs)
+        self.tag = "html"
+
+
+    def render(self, file_out, ind="", depth=1):
```
PythonCHB, line 74: yup!

students/mattmaeda/session07/html_render.py
```diff
@@ -0,0 +1,173 @@
...
+            file_out.write("{0}<{1}>{2}</{1}>\n".format(ind*depth, self.tag,
+                                                        elem))
+
+
+class SelfClosingTag(Element):
+    def __init__(self, content=None, **kwargs):
+        super(SelfClosingTag, self).__init__(content, **kwargs)
+
+
+    def render(self, file_out, ind = "", depth=1):
```
PythonCHB, line 62: is this going to handle attributes? or have you not done that yet?

students/mattmaeda/session07/test_html_render.py
```diff
@@ -0,0 +1,203 @@
...
+<tag>
+    Second line
+</tag>
+<tag>
+    Third line
+</tag>
+"""
+
+
+def test_init():
```
PythonCHB, line 32: non unittest tests! OK, I'll stop naggin now :-)

students/mattmaeda/session07/test_html_render.py
```diff
@@ -0,0 +1,203 @@
...
+    """
+    element = hr.Element("First line")
+    element.append("Second line")
+    element.append("Third line")
+
+    f = StringIO()
+    element.render(f, "")
+
+    f.seek(0)
+    assert f.read() == TEST_RENDER_NO_INDENT
```
PythonCHB, line 57: how could you test this without checking against a specific output?

students/mattmaeda/session07/test_html_render.py
```diff
@@ -0,0 +1,203 @@
...
+    assert f.read() == "<p>\n    This is my paragraph content\n</p>\n"
+
+
+def test_oneline_tag():
+    """ Test rendering of a one line tag element
+    """
+    one = hr.OneLineTag("This is my oneliner")
+
+    f = StringIO()
+    one.render(f, "    ")
```
PythonCHB, line 116: so are you using "depth" or the level of indentation here.

PythonCHB (approved):
nice work.

key point -- be careful with testing, it's easy to simply test exact results, but that makes the tests very fragile.

PR #114, opened 2017-11-10
Mailroom Test Question

brhelodrvr:
@PythonCHB @codedragon I'm using Chris's mailroom file as the basis of my test_mailroom.py script.  My mailroom script isn't working yet.  I get a `NameError` telling me that donor_db is not defined.  That makes sense since donor_db is defined elsewhere.  So,
1. How do I test the `list_donors()` function?  More generally, how do I test a function that uses elements like donor_db that are defined elsewhere?
2. Similarly, how do I test functions that have parameters -- e.g. `find_donor(name)` -- passed in?


students/brian_warn/session06/mailroom_from_chb.py
```diff
@@ -0,0 +1,230 @@
+#!/usr/bin/env python
```
brhelodrvr, line 1: @PythonCHB @codedragon Using Chris's code since my mailroom file isn't working yet.

students/brian_warn/session06/test_mailroom.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+def test_get_donor_db():
+    result = mr.get_donor_db()
+    assert result == {'william gates iii': ("William Gates III", [653772.32, 12.17]),
+        'jeff bezos': ("Jeff Bezos", [877.33]),
+        'paul allen': ("Paul Allen", [663.23, 43.87, 1.32]),
+        'mark zuckerberg': ("Mark Zuckerberg", [1663.23, 4300.87, 10432.0]),}
+
+def test_list_donors():
```
brhelodrvr, line 16: @PythonCHB @codedragon This fails with `NameError: name 'donor_db' is not defined` error.  I don't know how to test for functions that get content outside of the function itself.

```
___________________________________________________________________________ test_list_donors ___________________________________________________________________________

    def test_list_donors():
>       result = mr.list_donors()

test_mailroom.py:17: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

    def list_donors():
        """
        Create a list of the donors as a string, so they can be printed
    
        Not calling print from here makes it more flexible and easier to
        test
        """
        listing = ["Donor list:"]
>       for donor in donor_db.values():
E       NameError: name 'donor_db' is not defined

mailroom_from_chb.py:39: NameError
```

PythonCHB: Brian -- you are running into an anticipated issue!

Remember that I said something like "if your code is hard to test, that probably means it could use some refactoring!

In fact, I originally wrote mailroom without tests, and when I went to add the tests, I ended up having to refactor it a bit to make it more testable.

What you are running into now is one of the reasons why people generally frown on using "globals" -- it means that a function can't run properly outside of it's original context.

(it also means that you can't easily tell where things might get altered.)

I'll take a look at what you've got and see what I can suggest for refactoring -- either the code or your tests :-)

PythonCHB: OK -- found the solution.

Since the code IS in fact structured around a global donor_db, then you need to make sure it exists when you run your tests.

Note in the code, that the donor_db is created by a function --this is so that it would be fairly easy to trade-out code that, for instance, read the db from a file, or ...

So in the script, the donor_db is loaded into a global variable in the if __name__ block:

```
if __name__ == "__main__":
    donor_db = get_donor_db()
```

but, as you've noticed, that code doesn't get run when you run your tests!

So you need to run it by hand in your test file:

```
import mailroom2 as mailroom
mailroom.donor_db = mailroom.get_donor_db()
```
that creates the database, and sets it to a module-global variable.

Pretty wierd that you can set a module variable from outside the module -- but you can!

Ideally, that code would be in a "Fixture" -- but I don't want to complicate things for you too much, and using a global like that makes it hard to isolate anyway.

students/brian_warn/session06/test_mailroom.py
```diff
@@ -0,0 +1,18 @@
...
+ # test_mailroom.py
+
+import mailroom_from_chb as mr
+
+import pytest
+
+
+def test_get_donor_db():
+    result = mr.get_donor_db()
+    assert result == {'william gates iii': ("William Gates III", [653772.32, 12.17]),
```
PythonCHB, line 11: Careful with tests like this. In a way, you are not testing the logic in the function, but rather the data that's there. And that is just a test data set. If this code were moved into production, that data would not stay the same.

I think I chose not to test it at all -- if the rest of the tests pass, then there must be a valid donor_db object.

This actually gave me an idea. the get_donor_db function really is simply supplying a test dataset.

We needed it, so that we could test and demonstrate the code. But now that we're adding tests, it's probably better that the test data be embedded in the tests!

So you may want to copy the contents of the get_donor_db function into your tests:

```
# so that it's there for the tests
mailroom.donor_db = {'william gates iii': ("William Gates III", [653772.32, 12.17]),
                     'jeff bezos': ("Jeff Bezos", [877.33]),
                     'paul allen': ("Paul Allen", [663.23, 43.87, 1.32]),
                     'mark zuckerberg': ("Mark Zuckerberg", [1663.23, 4300.87, 10432.0]),
                     }
```
Now you have a test dataset to use, AND most importantly, you've defined the test dataset in the tests themselves - which is key, as you need to use the values in your tests.

And now you can change the sample data in mailroom itself, or change where it gets it, etc, without breaking your tests.

brhelodrvr: Thank you @PythonCHB (cc'ing @codedragon) for this helpful info.  This does indeed work -- assuming I'm implementing this test correctly.  Is the test as I've now defined it (below) the right way to perform this test?
```
def test_get_donor_db():
    mr.donor_db = {'william gates iii': ("William Gates III", [653772.32, 12.17]),
                     'jeff bezos': ("Jeff Bezos", [877.33]),
                     'paul allen': ("Paul Allen", [663.23, 43.87, 1.32]),
                     'mark zuckerberg': ("Mark Zuckerberg", [1663.23, 4300.87, 10432.0]),
                     }
    result = mr.get_donor_db()
    # assert result == {'william gates iii': ("William Gates III", [653772.32, 12.17]),
    #     'jeff bezos': ("Jeff Bezos", [877.33]),
    #     'paul allen': ("Paul Allen", [663.23, 43.87, 1.32]),
    #     'mark zuckerberg': ("Mark Zuckerberg", [1663.23, 4300.87, 10432.0]),}
    assert result == mr.donor_db
```

PythonCHB: actually, not quite :-)

in the line:
`mr.donor_db = {'william ga .... `

you are setting the donor_db name in the mailroom module.

then in the next lines:

```
result = mr.get_donor_db()`
assert result == mr.donor_db
```
You are checking that get_donor_db returns the same data.

But: you don't need to set `mr.get_donor_db` to test if you are getting what you want from `get_donor_db`. And youshoudn't, because that's a side effect of the test that you don't want -- the mailroom module has now been modified by this test, even though there was no need to.

So you could just change `mr.donr_db` to `donor_db` in two places, as you're good. 

Then there is the question of whether this is something to test.

mr.get_donor_db is providing a base donor database for the rest of the code to work with.

At this point, it's a hard-coded sample database. But if this code were ever deployed -- that sample data would be replaced with real data -- or an empty dataset, or...

so testing that the sample data is what you expect isn't really useful.

maybe I should have called the function: get_sample_donor_data() :-)

If you do test get_donor_db -- I'd test that you got a valid database -- it should be a dict, and if it isn't empty, the keys in the dict should be strings, and the values tuples with the first element a string, and the second a list of numbers.

But I'm not sure that's worth doing -- you want to test actual functionality -- and loading a big literal is expected to work :-)

And if the sample data is bad -- all the other tests are going to fail :-)

brhelodrvr: Thank you @PythonCHB for this response since I now see that I need to answer the simple "Should I test this function?" before setting up a test in the first place.  I had been proceeding with the approach that every function does require unit testing.

PythonCHB: Yup! The pedantic approach is that every function needs testing -- but in some cases it may not help :-)

I do sometimes put "do nothing" tests in, 'cause then you are at least testing that the function exists. So no harm in doing something simple like:

```
def test_get_donor_db()
    assert mr.get_donor_db is not None
```
it does test that the function exists, and that it does return _something_.

And it'll get your test coverage numbers up :-)

In fact, as I think about it, you DON'T want to test that a function like that returns exactly the test data you have at the moment -- now you've got all sorts of stuff to change if you change your test data set!

PythonCHB: oops, jsut saw a typo in that -- the do almost nothing test should be:

```
def test_get_donor_db()
    assert mr.get_donor_db() is not None
```

note that it is CALLING get_donor_db -- assuring you that it can be run :-)

PythonCHB:
I'm merging, so we'll know what's new when you are ready for review and do a new PR.

PR #115, opened 2017-11-10
refactored and partially tested mailroom

Eowyn42:
It is REALLY hard to test user input and printed-out strings. I refactored in a big way to test as many parts of code as I could, and get the three "while" loops to the point where they're all calling on other code to handle user input (not tested) and doing actions (tested). Now all my while loops look almost exactly alike except with different selections of actions and user prompts. 

py.test --cov=mailroom tells me I only have 60% test coverage so I need to figure out how to do things like, test if the right action was executed (I can test if no proper action was selected), test if the right list was written to the terminal, test if the report table pretty-prints properly. 

I'll be bringing this to office hours either saturday or sunday.

### DIFF
--- a/students/eowyn/session06/mailroom/mailroom.py
+++ b/students/eowyn/session06/mailroom/mailroom.py
@@ -1,0 +1,237 @@
+#!/usr/bin/env python
+
+import sys
+
+''' --------------------------------------------------------------
+GLOBAL VARIABLE: DONORS is the database of donor names, amounts
+-------------------------------------------------------------------'''
+donor_names = ["Margaret Atwood", "Fred Armisen",
+               "Heinz the Baron Krauss von Espy"]
+donations = [[300, 555], [240, 422, 1000], [1500, 2300]]
+DONORS = dict(zip(donor_names, donations))
+del donor_names
+del donations
+
+
+''' --------------------------------------------------------------
+Following are helper functions to control program flow
+-------------------------------------------------------------------'''
+
+
+def safe_input():
+    return None
+
+
+def quit_code():
+    sys.exit()
+
+
+def return_to_menu():
+    ''' Return True to trigger exit out of sub-loop'''
+    return True
+
+
+''' --------------------------------------------------------------
+Following are helper functions for generating reports
+-------------------------------------------------------------------'''
+
+
+def generate_report_data():
+    [name, total, number, ave] = [[], [], [], []]
+    for x in DONORS:  # loop over the keys which are donor names
+        name.append(x)
+        total.append(round(sum(DONORS[x])))
+        number.append(len(DONORS[x]))
+        ave.append(round(total[-1] / number[-1]))
+    report_data = list(zip(name, total, number, ave))
+    report_data = sorted(
+        report_data, key=lambda y: int(y[1]), reverse=True)
+    return report_data
+
+
+def generate_table():
+    ''' Pretty-print a table of donor statistics '''
+    print_table(generate_report_data())
+
+
+def setup_table():
+    ''' Set up the donor record table headers '''
+    headers = ('Donor Name', 'Total Given', 'Num Gifts', 'Average Gift')
+    fs1 = '|'.join(["{:<40}", "{:<12}", "{:<10}", "{:<12}"])
+    width = len(fs1.format(*headers))
+    head = (fs1.format(*headers)) + '\n'
+    head = head + (width * "-") + '\n'
+    return head
+
+
+def setup_body(list_data):
+    ''' Set up the donor records table body '''
+    body = ""
+    fs2 = ''.join(["{:<40}", "${:^12.2f}", "{:^12d}", "${:^12.2f}"])
+    for i in range(len(list_data)):
+        entry = list_data[i]
+        body = body + fs2.format(*entry) + '\n'
+    return body
+
+
+def print_table(list_data):
+    ''' Pretty-print the donor records '''
+    print(setup_table())
+    print(setup_body(list_data))
+
+
+''' --------------------------------------------------------------
+Following are helper functions for adding donations to DONORS and
+sending thank you notes to donors
+-------------------------------------------------------------------'''
+
+
+def collect_donor_input():
+    ''' Get name and donation amount to add to DONORS'''
+    fullname = input("Enter a donor name (new or existing):\n")
+    fullname = remove_inputquotes(fullname)
+    amount = float(input("Donation amount: "))
+    return (fullname, amount)
+
+
+def retrieve_donations(fullname):
+    ''' if donor exists, return donations, otherwise, return None'''
+    if fullname in DONORS:
+        return DONORS[fullname]
+
+
+def add_donation(fullname, amount):
+    ''' Update donor if it exists or add new donor if it does not'''
+    DONORS.setdefault(fullname, []).append(amount)
+
+
+def update_donors():
+    '''
+    Add new donation to DONORS; if new donor, add to DONOR.
+    Then print a thank-you note for the donation.
+    '''
+    (fullname, amount) = collect_donor_input()
+    add_donation(fullname, amount)
+    print(generate_letter(fullname))
+
+
+def list_donors():
+    ''' List all DONORS '''
+    print("All Donors:")
+    [print(x) for x in DONORS]
+
+
+def generate_letter(donor_name):
+    '''
+    Generate a formatted thank you note to a donor and return
+    the string of the thank you note for printing or writing.
+    '''
+    fs = "Thank you, {}, for your generosity and recent gift of ${}."
+    return fs.format(donor_name, DONORS[donor_name][-1])
+
+
+''' --------------------------------------------------------------
+Following are helper functions for accepting and responding to
+keyboard input
+-------------------------------------------------------------------'''
+
+
+def remove_inputquotes(a_string):
+    '''Remove auxillary quotes to cleanse a_string'''
+    a_string.replace('"', '')
+    return a_string
+
+
+def get_user_input(prompt_string):
+    ''' Print a prompt_string, return keyboard input if no exceptions'''
+    try:
+        answer = input(prompt_string)
+    except (EOFError, KeyboardInterrupt, TypeError):
+        return None
+    else:
+        return answer
+
+
+def select_action(arg_dict, answer):
+    ''' Execute an action from arg_dict that corresponds to answer.
+    Return None if action was executed and False if an error occurs'''
+    try:
+        return arg_dict.get(answer)()
+    except (KeyError, TypeError):
+        return False
+
+
+''' --------------------------------------------------------------
+Following are primary actions called by MAINLOOP
+-------------------------------------------------------------------'''
+
+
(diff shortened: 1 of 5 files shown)


PythonCHB:
Don't worry about testing the user interaction. The "right" way to do it is with mocking, or running test at a higher level that run it as a script. Maybe I"ll try to figure out how to mock input() -- but we'll probably wait 'till the second quarter when we get deeper into testing.

My solution has 57% coverage :-)

100% coverage is not the be-all and end-all -- testing the user interaction part is more integration testing anyway. Also, this is an odd case, as we really have a minimal amount of logic code, so it makes the coverage look bad...

students/eowyn/session06/mailroom/mailroom.py
```diff
@@ -0,0 +1,237 @@
...
+def quit_code():
+    sys.exit()
+
+
+def return_to_menu():
+    ''' Return True to trigger exit out of sub-loop'''
+    return True
+
+
+''' --------------------------------------------------------------
```
PythonCHB, line 34: you want to use comments, rather than triple-quoted strings for this sort of thing:

```
# Following are helper functions for generating reports
# -----------------------------------------------------
```

students/eowyn/session06/mailroom/test_mailroom.py
```diff
@@ -0,0 +1,122 @@
+#!/usr/bin/env python
+
+import mailroom
+import pytest
+import os.path
+
+
+def test_safeinput():
```
PythonCHB, line 8: do you really want this to return None?

again -- it's OK not to test user-interaction!

students/eowyn/session06/mailroom/test_mailroom.py
```diff
@@ -0,0 +1,122 @@
...
+    result = mailroom.safe_input()
+    assert result is None
+
+
+def test_return_to_menu():
+    result = mailroom.return_to_menu()
+    assert result is True
+
+
+def test_remove_inputquotes():
```
PythonCHB, line 18: good one.

students/eowyn/session06/mailroom/test_mailroom.py
```diff
@@ -0,0 +1,122 @@
...
+    astring = "'hello'"
+    output = 'hello'
+    assert mailroom.remove_inputquotes(astring) == output
+
+
+def test_generate_letter():
+    mailroom.DONORS = {'Fred Armisen': [240, 422, 1000],
+                       'Heinz the Baron Krauss von Espy': [1500, 2300],
+                       'Margaret Atwood': [300, 555]}
+    output = 'Thank you, Margaret Atwood, for your generosity and recent gift of $555.'
```
PythonCHB, line 28: be a bit careful about testing for exact results.

If you change the letter template a bit, it will fail --not too bad, you can update the test. Bu tyou may want to test if critical parts of the letter are there -- the right name, the right dollar amount, rather than all the text.

students/eowyn/session06/mailroom/test_mailroom.py
```diff
@@ -0,0 +1,122 @@
...
+    assert mailroom.generate_letter('Margaret Atwood') == output
+
+
+def test_generate_report_data():
```
PythonCHB, line 32: same here -- maybe test for critical components of the report -- correct number of entries, etc..

students/eowyn/session06/mailroom/test_mailroom.py
```diff
@@ -0,0 +1,122 @@
...
+                       'Heinz the Baron Krauss von Espy': [1500, 2300],
+                       'Margaret Atwood': [300, 555]}
+    output = [('Heinz the Baron Krauss von Espy', 3800, 2, 1900),
+              ('Fred Armisen', 1662, 3, 554),
+              ('Margaret Atwood', 855, 2, 428)]
+    assert mailroom.generate_report_data() == output
+
+
+def test_setup_table():
+    ''' this is not working'''
```
PythonCHB, line 43: what's wrong? maybe missing a linefeed at the end?

students/eowyn/session06/mailroom/test_mailroom.py
```diff
@@ -0,0 +1,122 @@
...
+    list_data = mailroom.generate_report_data()
+    output = """'Heinz the Baron Krauss von Espy         $  3800.00        2      $  1900.00   \nFred Armisen                            $  1662.00
+3      $   554.00   \nMargaret Atwood                         $   855.00        2      $   428.00   \n'
+"""
+    assert mailroom.setup_body(list_data) == output
+
+def test_remove_inputquotes():
+    output = 'hello'
+    assert mailroom.remove_inputquotes("hello") == output
```
PythonCHB, line 61: didn't you test this above?

Eowyn42:
Hey that's good to hear, maybe I'm done :-) My husband works in test
development for tableau so I'm hearing a lot of "hey you could mock up
input data such such and such ... end to end... " which at this point
sounds like a lot of gibberish so I'm more than happy to stop at ~60%
coverage on this one!

students/eowyn/session06/mailroom/test_mailroom.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+
+def test_retrieve_donor_exists():
+    mailroom.DONORS = {'Fred Armisen': [240, 422, 1000],
```
PythonCHB, line 66: by the way, this is great -- putting your test data in the test!

though if you're using teh same data in multiple tests, maybe put it in a variable -- but still set it here.

students/eowyn/session06/mailroom/test_mailroom.py
```diff
@@ -0,0 +1,122 @@
...
+    output = [300, 555]
+    assert mailroom.retrieve_donations(name) == output
+
+def test_retrieve_donations_new():
+    mailroom.DONORS = {'Fred Armisen': [240, 422, 1000],
+                       'Heinz the Baron Krauss von Espy': [1500, 2300],
+                       'Margaret Atwood': [300, 555]}
+    name = "Paul Simon"
+    output = None
+    assert mailroom.retrieve_donations(name) == output
```
PythonCHB, line 79: for something as simple as this, you can make it a one-liner:

 assert mailroom.retrieve_donations("Paul Simon") is None

and do use "is" for comparing to None

students/eowyn/session06/mailroom/test_mailroom.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+def test_add_donation_new():
+    mailroom.DONORS = {'Fred Armisen': [240, 422, 1000],
+                       'Heinz the Baron Krauss von Espy': [1500, 2300],
+                       'Margaret Atwood': [300, 555]}
+    name = "Bob Mueller"
+    amount = 500
+    mailroom.add_donation(name, amount)
+    assert mailroom.retrieve_donations(name) == [amount]
```
PythonCHB, line 89: perfect!

students/eowyn/session06/mailroom/test_mailroom.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+def test_add_donation_existing():
+    mailroom.DONORS = {'Fred Armisen': [240, 422, 1000],
+                       'Heinz the Baron Krauss von Espy': [1500, 2300],
+                       'Margaret Atwood': [300, 555]}
+    name = "Fred Armisen"
+    amount = 500
+    mailroom.add_donation(name, amount)
+    assert mailroom.retrieve_donations(name) == [240, 422, 1000, 500]
```
PythonCHB, line 99: this is fine, though it could be a tiny bit more general by testing for only the new amount:

`assert mailroom.retrieve_donations(name)[-1] == amount`

or, at least use the "amount" variable -- as long as you've defined it!

`assert mailroom.retrieve_donations(name) == [240, 422, 1000, amount]`

students/eowyn/session06/mailroom/test_mailroom.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+def test_select_action1():
+    ''' not sure how to test if correct action was completed
+    something called "mock" has this capability?'''
```
PythonCHB, line 104: yup --but that's complicated enough that wouldn't go there yet!

PythonCHB:
Well, you got me looking into how to mock input() -- turns out there are multiple ways:

https://stackoverflow.com/questions/35851323/pytest-how-to-test-a-function-with-input-call

I'm going to have to look deeper and decide what a "best practice" might be.

So I'd say you're done unless you REALLY want to figure it out now!

students/eowyn/session06/mailroom/test_mailroom.py
```diff
@@ -0,0 +1,122 @@
...
+def test_file_existance():
+    mailroom.DONORS = {'Fred Armisen': [240, 422, 1000],
+                       'Heinz the Baron Krauss von Espy': [1500, 2300],
+                       'Margaret Atwood': [300, 555]}
+    mailroom.send_letters()
+    c1 = os.path.isfile("Margaret_Atwood.txt")
+    c2 = os.path.isfile("Fred_Armisen.txt")
+    c3 = os.path.isfile("Heinz_the_Baron_Krauss_von_Espy.txt")
+    output = c1 and c2 and c3
+    assert output is True
```
PythonCHB, line 117: you can do this with multiple asserts:

assert os.path.isfile("Margaret_Atwood.txt")
assert os.path.isfile("Fred_Armisen.txt")
assert os.path.isfile("Heinz_the_Baron_Krauss_von_Espy.txt")

you could even put that in a loop to automate it

But be careful -- they could be left over from a previous test run!

don't worry about that for now though.

students/eowyn/session06/mailroom/mailroom.py
```diff
@@ -0,0 +1,237 @@
...
+    else:
+        return answer
+
+
+def select_action(arg_dict, answer):
+    ''' Execute an action from arg_dict that corresponds to answer.
+    Return None if action was executed and False if an error occurs'''
+    try:
+        return arg_dict.get(answer)()
+    except (KeyError, TypeError):
```
PythonCHB, line 160: If you  get a TypeError here it means the arg_dict is broken -- do you want to catch that?

PythonCHB (approved):
this looks great -- you've got the idea!

Eowyn42:
Hi Chris & Maria,

I had a great time in Alaska. It wasn’t too cold and the aurora came out
the last night :-) I’m working on getting caught up on coursework this
week, though we have tickets to see “Hamilton” today so unfortunately I
can’t make office hours. Life is pretty crazy right now but mostly in a
good way, like I’ve been trying to get Hamilton tickets for *years*.

I’m studying this page:
https://uwpce-pythoncert.github.io/PythonCertDevel/modules/MetaProgramming.html#metaprogramming

To do this assignment:
https://uwpce-pythoncert.github.io/PythonCertDevel/exercises/mailroom-meta.html#exercise-mailroom-meta

And will read the fluent python chapters (great book) and watch the
suggested YouTube video, with the goal of being prepared to move forward on
Tuesday.

Thanks for your understanding.

Best wishes,
Eowyn

PR #116, opened 2017-11-12
adding more files

zandra2:
I just added a bunch of files for you to review. I am not sure if this pull request applies to all of them. When adding files from Git. I didn't open the directory, but I provided the path from the directory to the file and add it from there. After I did that I remember that you said something like, Git is not very good at recognizing directory. So I am not sure if all the files got pushed correctly and if this pull request applies to all files. It looks like it all made?.  :-)

### DIFF
--- a/students/zandra_eng/session04/trigram.py
+++ b/students/zandra_eng/session04/trigram.py
@@ -1,0 +1,31 @@
+#!/usr/bin/env python
+import random
+
+infile = 'sherlock_small.txt'
+
+def removePunc():
+    """ removing punctuations and white space from lines of text and return list of remaining words"""
+    line_list = []
+    with open(infile, 'r') as fin:
+        for line in fin:
+            line = line.strip().replace('-', ' ')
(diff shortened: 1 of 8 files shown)


PythonCHB:
there is a lot here -- so I think you're good!

students/zandra_eng/session04/trigram.py
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env python
```
PythonCHB, line 1: You've got the basic logic working -- nice job!

students/zandra_eng/session05/except_exercise.py
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/python
```
PythonCHB, line 1: This a good to go through to get the hang of execption handling.

PythonCHB (approved):
good so far -- looks like there is lots more to add!

for the new stuff: 

git add the_name_of_the_file

git commit -a

git push

And to keep in sync with the class repo:

git pull upstream master

git push

PR #117, opened 2017-11-12
Pull Request

danwoj:
As you're well aware, I had some serious issues with my Git configuration so I've uploaded quite a lot. Now that I'm over that hurdle I'll be much better about committing my stuff on a more regular basis. In the interest of soliciting feedback from the both of you without having to go through everything I've committed, I've picked a few programs, listed below, that I'd appreciate you taking the time to review for me.

danwoj/Session03/
rot13.py

danwoj/Session04/
mailroom_original.py (my original mail room program)
trigrams.py

danwoj/Session05/
mailroom_dict.py (mail room with dictionaries)


PythonCHB:
Thanks for the note -- that's a LOT of files!

I'll take a look at the ones you suggested :-)

students/danwoj/Session03/rot13.py
```diff
@@ -0,0 +1,42 @@
+def rot13(string):
+	cipher = ''
+# This function begins by going through each character and determining its Unicode value
+	for i in range(len(string)):
```
PythonCHB, line 4: you can loop through teh charaters with:

``
for c in string:
    ...
``

students/danwoj/Session03/rot13.py
```diff
@@ -0,0 +1,42 @@
...
+		y = ord(string[i:i+1])
+		# This conditional is triggered if a capital letter is identified
+		if y >= 65 and y <= 90:
+			# This conditional converts the letter to ROT13
+			if (y+13) <= 90:
```
PythonCHB, line 9: a little more "clean" to add the 13, then check if it's too big.

students/danwoj/Session03/rot13.py
```diff
@@ -0,0 +1,42 @@
+def rot13(string):
```
PythonCHB, line 1: You seem to have tabs for indentation -- be sure to set your editor to use four spaces for indentation, even when you hit the <tab> key.

And it should be able to convert an entire file for you all at once.

students/danwoj/Session03/rot13.py
```diff
@@ -0,0 +1,42 @@
...
+			if (y+13) <= 122:
+				z = chr(y+13)
+				cipher = cipher + z
+			# This else statement essentially 'resets' the ROT13 substitution if the letter goes over the bounds of lowercases in Unicode
+			else:
+				z = chr(y-13)
+				cipher = cipher + z
+		# This final else statement ignores any characters that are neither capital nor lowercase letters. This allows for puncuation to persist in the ROT13 substitution
+		else:
+			cipher = cipher + string[i:i+1]
```
PythonCHB, line 28: strings are kind of a special case --indexing into a string gets you a single character string, so:

string[i] is exactly the same as string[1:1+1]

but you really want one character, so simple indexing makes more sense here:

students/danwoj/Session03/rot13.py
```diff
@@ -0,0 +1,42 @@
...
+	return cipher
+
+def mainloop():
```
PythonCHB, line 31: this isn't a loop, so you probably don't want to call it mainloop -- plain old main() is a traditional name for the main function.

students/danwoj/Session03/rot13.py
```diff
@@ -0,0 +1,42 @@
...
+	i_value2 = 'Magnetic from outside near corner!'
+
+	assert rot13(i_value1) == 'Magnetic from outside near corner.'
+	assert rot13(i_value2) == 'Zntargvp sebz bhgfvqr arne pbeare!'
+
+
+if __name__ == '__main__':
+	mainloop()
+
+
```
PythonCHB, line 42: could be a bit more compact, but you got the logic all correct, and the basic idea of working with ord() and chr() down.

students/danwoj/Session04/Prototypes/trigrams.py
```diff
@@ -0,0 +1,51 @@
...
+words = 'I wish I may I wish I might'.split()
+
+print(words)
+
+trigrams = {}
+for i in range(len(words)-2):
+	pair = tuple(words[i:i+2])
+	follower = words[i+2]
+	print('Pair and follower: ', pair, follower)
+	if pair in trigrams:
```
PythonCHB, line 12: perfect logic, but do check out dict.setdfault() for a more compact way to do it.

students/danwoj/Session04/Prototypes/trigrams2.py
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env python
```
PythonCHB, line 1: hmm, looks like this is the one I should be reviewing...

students/danwoj/Session04/Prototypes/trigrams2.py
```diff
@@ -0,0 +1,53 @@
...
+for i in range(len(new_words)-2):
+	pair = tuple(new_words[i:i+2])
+	follower = new_words[i+2]
+#	print('Pair and follower: ', pair, follower)
+	if pair in trigrams:
+		trigrams[pair].append(follower)
+	else:
+		trigrams[pair] = [follower]
+
+counter = (len(trigrams.keys()))//3
```
PythonCHB, line 39: why divided by 3?

students/danwoj/Session04/Prototypes/trigrams2.py
```diff
@@ -0,0 +1,53 @@
...
+
+print('Counter: ', counter)
+
+#print('This is trigrams: ', trigrams)
+
+import random
+
+for i in range(counter):
+	which_list, item = random.choice([(name, value) for name, values in trigrams.items() for value in values])
```
PythonCHB, line 48: slick logic!

though perhaps a bit heavy-weight --it's going to create a full list of all possible pair, follower combinations.

and now that i think about, a bunch that aren't valid, as you get followers that don't belong with the pairs!

and re-create that every time in the loop!

Better to start with just a random pair:  random.choice(list(trigrams))

then pick a random follower that matches that pair.

students/danwoj/Session04/Prototypes/trigrams2.py
```diff
@@ -0,0 +1,53 @@
...
+	pair_word1, pair_word2 = which_list
+	#print('which_list: ', which_list)
+	#print('item: ', item)
+	print(pair_word1, pair_word2, item, '', end='')
```
PythonCHB, line 52: so this isn't wuite the rigth logic. You want to start with a pair, and then add one of the workds that followed that pair in the text.

You know have three words. so the last two are the new pair, and you can add a follower to that, giving you four words. the last two are now a new pair.... etc.

students/danwoj/Session04/Prototypes/mailroom_dict.py
```diff
@@ -0,0 +1,108 @@
+#!/usr/bin/env python3.6
```
PythonCHB, line 1: I'm just going to review this version -- I assume it's the "latest and greatest" :-)

students/danwoj/Session04/Prototypes/mailroom_dict.py
```diff
@@ -0,0 +1,108 @@
...
+
+# This creates a statis donor table that does not retain values after this program runs
+
+donor_fnames = ['Bill', 'Melvin', 'Daphnie', 'Chauncey', 'Frieda']
+donor_lnames = ['Gates', 'Smith', 'Jones', 'Doe', 'Whatever']
+donations = [[200, 500], [2000.00, 3000.50, 3000.00], [1500.00, 500.25], [400.00], [2000.12, 3000, 4000]]
+
+keys = ['f_name', 'l_name', 'don_amt', 'don_num']
+donor_dict = dict.fromkeys(keys, None)
+donor_dict['f_name'] = donor_fnames
```
PythonCHB, line 11: this is over-righting what you had in the dict above -- no need for fromkeys() is you are goignt o set the real keys and values later.

students/danwoj/Session04/Prototypes/mailroom_dict.py
```diff
@@ -0,0 +1,108 @@
...
+donor_dict['l_name'] = donor_lnames
+donor_dict['don_amt'] = donations
+donor_dict['don_num'] = []
+
```
PythonCHB, line 15: this is not really a great data structure to have -- you've not separated the various parts of a donor from each other -- you can only link them by index, which is pretty prone to error.

Better to have a data structure that keeps everything that has to do with one donor together -- name, dontation aounts, etc... -- this can be a simple tuple:

(name, donat_list)

or a dict:
```
{"first_name": "fred",
 "last_name": "Jones",
"donations", [345, 1000, ]
}

then if you want to store the whole thing a dict -- good for easy lookup, put each donor "record" in a dict, with maybe the name as the key (or a normalized version of the name -- no spaces, all lower case, or..).

File: students/danwoj/Session04/Prototypes/mailroom_dict.py [hunk 15]
```diff
@@ -0,0 +1,108 @@
...
+for i in donor_dict['don_amt']:
+	donor_dict['don_num'].append(len(i))
+
+# This is the function that prints off the donor report
+def print_report():
```
PythonCHB, line 20: this is where it would be good to split out the report generation from the printing -- have this runction return a string to print, but actually print it elsewhere.

when you try to test it -- you'll see why!

students/danwoj/Session04/Prototypes/mailroom_dict.py
```diff
@@ -0,0 +1,108 @@
...
+	print('\n#####################  Donor Report ####################\n')
+	print('Donor Name      | Total Given | Num Gifts | Average Gift')
+	print('--------------------------------------------------------')
+	i = 0
+	# This loop goes through each donor record in the dataset
+	for i in range (len(donor_names)):
+		j = 0
+		tot_don = 0
+		avg_don = 0
+		# This loop creates the total and average donation values for each donor
```
PythonCHB, line 30: you may also want to have calculating the average donations separate from the report creation.

students/danwoj/Session04/Prototypes/mailroom_dict.py
```diff
@@ -0,0 +1,108 @@
...
+
+# This is the function that prints off the donor report
+def print_report():
+	print('\n#####################  Donor Report ####################\n')
+	print('Donor Name      | Total Given | Num Gifts | Average Gift')
+	print('--------------------------------------------------------')
+	i = 0
+	# This loop goes through each donor record in the dataset
+	for i in range (len(donor_names)):
+		j = 0
```
PythonCHB, line 27: see how you are having to keep an index to keep the "parts" of donors' information in sync?

students/danwoj/Session04/Prototypes/mailroom_dict.py
```diff
@@ -0,0 +1,108 @@
...
+		for j in range (len(donations[i])):
+			tot_don = tot_don + donations[i][j]
+			avg_don = avg_don + donations[i][j]
+		avg_don = avg_don / (j+1)
+		# This prints off each donor's row in the report
+		print(donor_names[i], ' '*(14-len(donor_names[i])), '| $', '%.2f' % tot_don, ' '*(8-len('%.2f' % tot_don)), '|', j+1, ' '*(8-len(str(j+1))), '| $', '%.2f' % avg_don)
+	return True
+
+# This function prints out the 'thank you' message for a new donation
+def ty_message(dn, da):
```
PythonCHB, line 40: again good to generate the letter, then print it elsewhere.

students/danwoj/Session04/Prototypes/mailroom_dict.py
```diff
@@ -0,0 +1,108 @@
...
+	tym_value = int(input('\n#####################\n#  Send Thank You   #\n#####################\n'
+		'1: List Donors\n'
+		'2: Enter Donation\n'
+		'3: Go Back\n'
+		'Enter your number choice: '))
+	# Creates a donor list if the user selects that option from the menu
+	if tym_value == 1:
+		i = 0
+		print('\nDONOR LIST:')
+		for i in range (len(donor_names)):
```
PythonCHB, line 57: better to put this in a separate function as well.

there should be as little logic code as possible in the interactive loop.

students/danwoj/Session04/Prototypes/mailroom_dict.py
```diff
@@ -0,0 +1,108 @@
...
+			print(donor_names[i])
+		thank_you()
+	# Enters a new donation if the user selects that option from the menu
+	elif tym_value == 2:
+		i = 0
+		new_don = 1
+		d_name = input("Please enter the donor's name: \n")
+		# Loop searches for the donor's name in the existing data set
+		for i in range (len(donor_names)):
```
PythonCHB, line 66: you see the logic requires to keep information in sync!

and another good place to put the logic in a function outside the interactive loop.

PythonCHB (approved):
got all the logic good, but mailroom could really use some refactoring.

Next week, we'll be looking at using classes to refactor it -- so take that oppurtunity to clean up the data structures and logic.

PR #118, opened 2017-11-12
Mailroom with full test coverage! Including def run_interactive_loop

Eowyn42:
No feedback needed -- was working on this during Sunday office hours. Responded to earlier feedback on pull request #115, and added a few more tests and a little more refactoring: specifically, putting the while loops into a run_interactive_loop(arg_dict, prompt_string) function which by itself increased by test coverage to 68%. 

to generate html of test coverage: 
pytest --cov=mailroom --cov-report=html
Only print statements, user input, and python functionality like "while" loops are untested :-)

### DIFF
--- a/students/eowyn/session06/mailroom/mailroom.py
+++ b/students/eowyn/session06/mailroom/mailroom.py
@@ -2,9 +2,10 @@
 
 import sys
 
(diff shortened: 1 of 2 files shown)


PythonCHB:
Awesome!

PR #119, opened 2017-11-12
Session03 code

AmitUW:


students/Amit/Session04/get_lang.py
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env python3
+from collections import defaultdict
+
+def list_langs():
+
+    with open('students.txt', 'r') as file:
+        file.readline()
+        languages = [lang.capitalize() for line in file for lang in line.strip().split(':')[1].replace(",", " ").strip().split() if not lang[0].isupper()]
```
PythonCHB, line 8: slick! 

Though maybe better to split this out to a couple lines -- it's pretty hard to see what's going on with a line this long.

students/Amit/Session04/get_lang.py
```diff
@@ -0,0 +1,30 @@
...
+def main():
+    dict_lang = {}
+    for language in list_langs():
+        if language not in dict_lang.keys():
+            dict_lang[language] = 0
+        dict_lang[language] += 1
+    return dict_lang
+
+def main2():
+    dict_lang = defaultdict(int)
```
PythonCHB, line 21: Nice find.

Though check out collections.counter -- even easier for this use case.

PythonCHB (approved):
mostly older stuff, so I just gave it a quick look-over -- let us now if you have anything in particular you want reviewed.

PR #120, opened 2017-11-13
johnn session06

johnnzz:
* switched from list of dicts to dict of dicts
* switched from index passing to dict passing
* added unit tests


johnnzz:
closing pull request to re-generate diff in order to pick up new changes

PR #122, opened 2017-11-13
johnn session06

johnnzz:
* switched from list of dicts to dict of dicts
* switched from index passing to dict passing
* added update dict function
* other minor refactoring to facilitate unit testing
* added unit tests

### DIFF
--- a/students/john_navitsky/session06/mailroom.py
+++ b/students/john_navitsky/session06/mailroom.py
@@ -69,16 +69,16 @@ def print_lines(lines=2,dest=sys.stdout):
         print("",file=dest)
 
 
-def print_report():
+def print_report(dest=sys.stdout):
     """ Print formatted list of all donors. """
 
-    print_lines()
+    print_lines(2,dest)
   
     # print report header
     #      index    name   total gvn #gifts  avg gift
     print("{0:5} | {1:20} | {2:14} | {3:9} | {4:12}".format(
-        "ID","Donor Name","Total Given","Num Gifts","Average Gift"))
-    print("-"*72)
+        "ID","Donor Name","Total Given","Num Gifts","Average Gift"),file=dest)
+    print("-"*72,file=dest)
 
     # enumerate so we have the list index handy
     for donor_index, donor in enumerate(donors):
@@ -93,8 +93,8 @@ def print_report():
 
         #      index      name     total gvn    #gifts    avg gift
         print("{0:05d}   {1:20}  ${2:14,.2f}   {3:9.0f}  ${4:12,.2f}".format(
-            donor_index,donor["first_name"]+" "+donor["last_name"],
-            total_donations,num_donations,average_donation))
+            donor_index,donor["full_name"],
+            total_donations,num_donations,average_donation),file=dest)
 
 
 def parse_name(full_name):
@@ -117,22 +117,22 @@ def parse_name(full_name):
 
         # ensure standard capitalization in stuffix
         if suffix:
-        	new_full_name=", ".join([informal_name, suffix])
+            new_full_name=", ".join([informal_name, suffix])
         else:
-        	new_full_name=informal_name
+            new_full_name=informal_name
 
         return { "full_name": new_full_name, "informal_name": informal_name,
             "suffix": suffix, "last_name": last_name, "first_name": first_name }
 
 
-def get_donation_amount(informal_name):
+def get_donation_amount(current_donor):
     """ Get a donation. """
 
     menu =  "\n"
     menu += "GIFT ENTRY (ammount)\n"
     menu += "-----------------------\n"
     menu += "\n"
-    menu += "Enter the numeric amount {} has donated or\n"
+    menu += "Enter the numeric amount {informal_name} has donated or\n"
     menu += "(q)uit to return to cancel the donation and\n"
     menu += "return to the previous menu.\n"
     menu += "\n"
@@ -142,7 +142,7 @@ def get_donation_amount(informal_name):
 
         print_lines()
 
-        print(menu.format(informal_name))
+        print(menu.format(**current_donor))
         selection=safe_input("Donation amount or (q)uit: ")
 
         # let them bail if they want
@@ -158,11 +158,11 @@ def get_donation_amount(informal_name):
             print("The value must be numeric.  Please try again or (q)uit.")
 
 
-def print_thank_you(donor_id,hint="wonderful",dest=sys.stdout):
+def print_thank_you(current_donor,hint="wonderful",dest=sys.stdout):
     """ Print a thank you letter. """
 
     letter =  "\n"
-    letter += "Dearest {first_name} {last_name},\n"
+    letter += "Dearest {full_name},\n"
     letter += "\n"
     letter += "We are are grateful for the generious donation on the behalf of\n"
     letter += "the {last_name} family.\n"
@@ -178,7 +178,7 @@ def print_thank_you(donor_id,hint="wonderful",dest=sys.stdout):
     print_lines(2,dest)
     print("-"*80,file=dest)
 
-    print(letter.format(hint,**donors[donor_id]),file=dest)
+    print(letter.format(hint,**current_donor),file=dest)
 
     print("-"*80,file=dest)
     print_lines(2,dest)
@@ -213,8 +213,16 @@ def thank_you_entry():
             print_report()
             continue
 
+        # protect against no entry
+        if not selection:
+            continue
+
         # parse the string into name components
         current_donor=parse_name(selection)
+
+        if current_donor["first_name"] == "" or current_donor["last_name"] == "":
+            print("You must enter both a first and last name.")
+            continue
 
         # determine if the provided name matches an existing record
         donor_id=None
@@ -223,12 +231,12 @@ def thank_you_entry():
             existing_full_name = existing_donor["first_name"]+" "+existing_donor["last_name"]
             # if we get a match, store the index of the donor record
             if existing_full_name.lower().strip() == current_donor["full_name"].lower().strip():
-            	current_donor=existing_donor
-            	#print("CUR:", current_donor)
-            	donor_id=donor_index
+                current_donor=existing_donor
+                #print("CUR:", current_donor)
+                donor_id=donor_index
 
         # prompt for new donation, cancel if None returned
-        new_donation=get_donation_amount(current_donor["informal_name"])
+        new_donation=get_donation_amount(current_donor)
         if new_donation is None:
             print_lines()
             print("Donation cancelled!")
@@ -261,17 +269,17 @@ def thank_you_entry():
             hint="new"
 
         # thank the donor for the new donation
-        print_thank_you(donor_id,hint)
+        print_thank_you(current_donor,hint)
 
 
 def thank_all_donors():
+    """ loop through donors, open file, print a thank you letter. """
     for index in range(len(donors)):
-        #dest=sys.stdout
-        donor_name = donors[index]["last_name"]+"_"+donors[index]["first_name"]
-        donor_name = donor_name.strip().lower().replace(" ", "_").replace(",", "")
-        donor_file="thank_you_" + donor_name
-        dest = open(donor_file, "w")
-        print_thank_you(index,"wonderful",dest)
+        donor=donors[index]
+        file_name="_".join(filter( None, [ "thank_you", donor["last_name"], donor["suffix"], donor["first_name"] ])).lower()
+        file_name=file_name.replace(" ", "_")
+        dest = open(file_name, "w")
+        print_thank_you(donor,"wonderful",dest)
         dest.close()
 
 
--- a/students/john_navitsky/session06/mailroom.py
+++ b/students/john_navitsky/session06/mailroom.py
@@ -189,6 +189,7 @@ def print_thank_you(current_donor,hint="wonderful",dest=sys.stdout):
 
 
 def match_donor(current_donor, donors):
+        """ See if record matches existing record, if so, return full record. """
         for donor_id in donors:
             existing_donor=donors[donor_id]
             if existing_donor["full_name"].lower() == current_donor["full_name"].lower():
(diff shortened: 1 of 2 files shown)


students/john_navitsky/session06/mailroom.py
```diff
@@ -20,9 +23,9 @@ def load_donor_file(donor_file="donors.json"):
             return donors
     except (FileNotFoundError):
             # if the file isn't found, that's ok, give them
-            # an empty donor list and we'll save anything
+            # an empty donor dict and we'll save anything
             # they enter upon exit.
-            return []
+            return {}
     except (PermissionError):
         # if the file is there, but you can't read it, don't continue
```
PythonCHB, line 30: nice! -- though a touch tricky -- it could have read-only permissions, then you'd be able to read it, so not get this error, but not be able to write on exit.

students/john_navitsky/session06/mailroom.py
```diff
@@ -67,19 +70,20 @@ def print_lines(lines=2,dest=sys.stdout):
         print("",file=dest)
 
 
-def print_report():
+def print_report(donors,dest=sys.stdout):
```
PythonCHB, line 73: Nice!

However, makes it more testable if you have a function that generates the report, and then print it  separately.

students/john_navitsky/session06/mailroom.py
```diff
@@ -67,19 +70,20 @@ def print_lines(lines=2,dest=sys.stdout):
...
     """ Print formatted list of all donors. """
 
-    print_lines()
+    print_lines(2,dest)
```
PythonCHB, line 76: also, if you want to be able to write to a file-like object, you might as well use the write method:

dest.write("-"*72)

sys.stdout.write will be the console -- so just like print (except it doesn't put newlines in for you).

students/john_navitsky/session06/mailroom.py
```diff
@@ -168,13 +182,52 @@ def print_thank_you(donor_id,hint="wonderful",dest=sys.stdout):
...
 
     print("-"*80,file=dest)
     print_lines(2,dest)
 
 
-def thank_you_entry():
+def match_donor(current_donor, donors):
+        """ See if record matches existing record, if so, return full record. """
+        for donor_id in donors:
+            existing_donor=donors[donor_id]
```
PythonCHB, line 194: If you want to loop through the values, you just do:

``for existing_donor in donors.values():``

students/john_navitsky/session06/mailroom.py
```diff
@@ -168,13 +182,52 @@ def print_thank_you(donor_id,hint="wonderful",dest=sys.stdout):
...
+            if existing_donor["full_name"].lower() == current_donor["full_name"].lower():
```
PythonCHB, line 195: having to loop though the whole thing to find a particular donor pretty much defeats the purpose of storing it all in a dict.

But look what you are doing -- you are considering a match if the full_name, all lower cased, is the same. So you could use the lower-cased version of the name as you dict keys --then easy lookup.

Take a look at my solution -- I did a "normalized" version of the name as the key.

students/john_navitsky/session06/mailroom.py
```diff
@@ -168,13 +182,52 @@ def print_thank_you(donor_id,hint="wonderful",dest=sys.stdout):
...
+        """ Update the donor database with new donation or donor record as needed. """
+
+        # match to existing record, if exists
+        current_donor = match_donor(current_donor, donors)
+
+        # datetime.datetime.strptime(<iso date>,'%Y-%m-%dZ')
+        today = datetime.datetime.utcnow().date().isoformat() + "Z"
+
+        # add donation to an existing donor
+        if "donor_id" in current_donor.keys():
```
PythonCHB, line 211: This is just `if donor_id in current_donor`

or handle this with:

```
try:
    donor_id = current_donor["donor_id"]
except KeyError
```

But why would you have a donor record without an id anyway? Better to always put in id in why you first create the donor record.

students/john_navitsky/session06/mailroom.py
```diff
@@ -168,13 +182,52 @@ def print_thank_you(donor_id,hint="wonderful",dest=sys.stdout):
...
+            donor_id=current_donor["donor_id"]
+            donors[donor_id]["donations"].append({ "amount": new_donation, "date": today })
+            hint="returning"
+        else:
+            # datetime.datetime.strptime(<iso time>,'%Y-%m-%dT%H:%M:%S.%fZ')
+            now = datetime.datetime.utcnow().isoformat() + "Z"
+            # id is a hash of the full donor name plus high resolution time
```
PythonCHB, line 218: if you are trying to simply get a unique id --maybe use uuid:

import uuid.

A couple other notes:

What function does an id serve here anyway?

and if you have an id in a donor record, why not simiply call it "id" -- if it's part of a donor record, we know it's a donor id.

students/john_navitsky/session06/mailroom.py
```diff
@@ -168,13 +182,52 @@ def print_thank_you(donor_id,hint="wonderful",dest=sys.stdout):
...
 
 
-def thank_you_entry():
+def match_donor(current_donor, donors):
+        """ See if record matches existing record, if so, return full record. """
+        for donor_id in donors:
+            existing_donor=donors[donor_id]
+            if existing_donor["full_name"].lower() == current_donor["full_name"].lower():
+                # if the owner exists, pull in the complete donor record
+                current_donor=existing_donor
```
PythonCHB, line 197: this is a bit of odd logic -- you are always going to get a donor record back, but it may be the new one, or it may be an existing one -- maybe better to search the donor db for a name, and then you can return either the existing one, or a new one, rather than not being sure at this point which it is.

students/john_navitsky/session06/mailroom.py
```diff
@@ -168,13 +182,52 @@ def print_thank_you(donor_id,hint="wonderful",dest=sys.stdout):
...
+            now = datetime.datetime.utcnow().isoformat() + "Z"
+            # id is a hash of the full donor name plus high resolution time
+            id_seed = current_donor["full_name"] + now
+            donor_id = hashlib.md5( id_seed.encode() ).hexdigest()
+            # add the donor
+            donors[donor_id] = { 
+                "created": now,
+                "donor_id": donor_id,
+                "donations": [ { "amount": new_donation, "date": today } ],
+                **current_donor }
```
PythonCHB, line 226: slick! I didn't realize you could do this.

students/john_navitsky/session06/mailroom.py
```diff
@@ -168,13 +182,52 @@ def print_thank_you(donor_id,hint="wonderful",dest=sys.stdout):
...
+        """ See if record matches existing record, if so, return full record. """
+        for donor_id in donors:
+            existing_donor=donors[donor_id]
+            if existing_donor["full_name"].lower() == current_donor["full_name"].lower():
+                # if the owner exists, pull in the complete donor record
+                current_donor=existing_donor
+        return current_donor
+
+
+def update_donor(current_donor, donors, new_donation):
```
PythonCHB, line 201: the logic here is a bit unclean. Probably better to do somethign like:

check if the dor is already there.

if it is, get that donor record 

if it isn't, create a new complete donor record and pass that along.

you now have a donor record to work with, and the rest of the code doesn't need to know if it's new or not.

students/john_navitsky/session06/mailroom.py
```diff
@@ -200,62 +253,51 @@ def thank_you_entry():
...
-        # determine if the provided name matches an existing record
-        donor_id=None
-        for donor_index, existing_donor in enumerate(donors):
-            # assemble full name from first_name, last_name
-            existing_full_name = existing_donor["first_name"]+" "+existing_donor["last_name"]
-            # if we get a match, store the index of the donor record
-            if existing_full_name.lower().strip() == current_donor["full_name"].lower().strip():
-                donor_id=donor_index
+        # reject blatantly bad input
+        if current_donor["first_name"] == "" or current_donor["last_name"] == "":
```
PythonCHB, line 267: since you are parsing the name elsewhere, best to put the validation and parsing all in one place.

students/john_navitsky/session06/mailroom.py
```diff
@@ -200,62 +253,51 @@ def thank_you_entry():
...
-        dest = open(donor_file, "w")
-        print_thank_you(index,"wonderful",dest)
-        dest.close()
+def thank_all_donors(donors,dest_override=None):
+    """ loop through donors, open file, print a thank you letter. """
+    for donor_id in donors:
+        donor=donors[donor_id]
+        file_name="_".join(filter( None, [ "thank_you", donor["last_name"], donor["suffix"], donor["first_name"] ])).lower()
+        file_name=file_name.replace(" ", "_")
+        if not dest_override:
```
PythonCHB, line 291: better to check for None explicitly.

`if dest_override is None:`

students/john_navitsky/session06/test_mailroom.py
```diff
@@ -32,12 +34,206 @@ def test_mock_input():
...
-#     assert "Dearest Joe Smith," in output
+def test_print_thank_you():
+    # verify the print_thank_you() function
+    a_donor = {
+        'full_name': 'Mary Jo Smith, IV', 
+        'informal_name': 'Mary Jo Smith',
+        'suffix': 'IV',
+        'last_name': 'Smith',
+        'first_name': 'Mary Jo' }
+    out = io.StringIO()
```
PythonCHB, line 45: nice use of StringIO.

students/john_navitsky/session06/test_mailroom.py
```diff
@@ -32,12 +34,206 @@ def test_mock_input():
...
-# def test_print_thank_you():
-#      # verify the print_thank_you() function
-#     out = io.StringIO()
-#     mailroom.print_thank_you(0,"testfull",out)
-#     output = out.getvalue()
-#     out.close()
-#     assert "Dearest Joe Smith," in output
+def test_print_thank_you():
+    # verify the print_thank_you() function
+    a_donor = {
```
PythonCHB, line 39: good use of donor example in test.

but shouldn't there be a donations entry too?

PythonCHB (approved):
really nice job with the well structured data.

And your tests look great.

students/john_navitsky/session06/mailroom.py
johnnzz: I did not want to use the name as a dict key because it is perfectly reasonable for there to be multiple donors with the exact same name which is why I previously choose to make a list of dicts*.  

However, while the list of dicts allows for reasonable representation of the data, I felt the update paradigm could use some improvement.  Updates invariably come down to use of the list index which seems open to concurrency issues and bugs are more likely to result in silent data corruption.

This led me to choose to create a unique identifier for each record and the use of it as a dict key,  While it is true this doesn't give the same lookup benefits of using the name as a key, I was less concerned with that particular benefit of dicts.

* Admittedly, I did not add the logic to actually allow the entry of duplicate donors of the same name at this time.

students/john_navitsky/session06/mailroom.py
johnnzz: Yah, I think you mentioned this previously, but I didn't have time to get around to changing.

students/john_navitsky/session06/mailroom.py
johnnzz: At some point you're going to need to decide if a record is new or matches an existing record.  Personally, I'd prefer this to happen when they start to enter a record.  This particular function is a really a stub where it should be expanded to allow them to select an existing entry or determine it is a new donor.

students/john_navitsky/session06/mailroom.py
johnnzz: Yah, it would probably be better to separate the donor record creation from the record update.

students/john_navitsky/session06/mailroom.py
johnnzz: This was a hold-over from when I had records from before the refactor so I was handling "updating" the contents.  At some point I decided migrating data wasn't a meaningful feature and I should have removed this.

That said, yes, I agree, I should make sure I create the id as soon as I know it isn't an existing record.

students/john_navitsky/session06/mailroom.py
johnnzz: Yes, I found UUID and went back and forth about which approach.  In the end I think I picked a hash mostly because it was a bit shorter and should serve same function.

I figured having the id as the key and in the record might raise an eyebrow.  In general I found it was a convenience.

Good point on the redundancy.

PythonCHB:
> I did not want to use the name as a dict key because it is perfectly
> reasonable for there to be multiple donors with the exact same name
>

Indeed it is :-)

which is
>
> However, while the list of dicts allows for reasonable representation of
> the data, I felt the update paradigm could use some improvement. Updates
> invariably come down to use of the list index which seems open to
> concurrency issues and bugs are more likely to result in silent data
> corruption.
>

Hmm, i’m Not sure that’s an issue, as you should be working with a
reference to the donor record, rather than in index anyway.

This led me to choose to create a unique identifier for each record and the
> use of it as a dict key,
>

Which means the Id could get out of sync with the key.

But anyway, you are thinking about this the right way, but writing a robust
database is getting a bit beyond the requirements of the assignment :-)


-- 
Christopher Barker, PhD

Python Language Consulting
  - Teaching
  - Scientific Software Development
  - Desktop GUI and Web Development
  - wxPython, numpy, scipy, Cython

students/john_navitsky/session06/mailroom.py
johnnzz: As a general application of "Explicit is better than implicit.", or is there a specific pitfall?  The reason I ask is I tend to really like the ability to treat objects as a Boolean.

students/john_navitsky/session06/test_mailroom.py
johnnzz: I probably overlooked it in the lesson description.

students/john_navitsky/session06/mailroom.py
PythonCHB: Yup — bit I think better to be more direct. 

And you don’t want to create incomplete records.

students/john_navitsky/session06/mailroom.py
PythonCHB: None is kind of a special case — I like objects as Boolean too, but None is one thing, rather than a value that happens to be a Falsey value.

If you don’t test for None, then 0 or an empty string or and empty sequence will also be false.

students/john_navitsky/session06/test_mailroom.py
PythonCHB: You’ve got it elsewhere ....

PythonCHB: You’ve got it elsewhere ....

PR #123, opened 2017-11-14
test file for mailroom.py

ea8606:
Not quite done but done enough for  a review


students/EricAdams/session06/test_mailroom.py
```diff
@@ -0,0 +1,221 @@
...
+
+
+def test_gen_letter():
+    '''
+    Input a valid donor and assert that the letter is returned
+    '''
+    letter = dedent('''
+          Dear {}
+
+          Thank you for your very kind donation of ${:.2f}.
```
PythonCHB, line 88: careful here -- this works, and is fine for now, and is helpful in you'll catch it if the function totally fails.

But if anything changes about the letter format, this will then fail, and you'll need to update the test to match the code, which is kind of a pain.

So ideally you test what is critical about the result, rather than that it's exactly what it is at the moment.

For instance: did the donors name get properly put in the letter?
Did the donation amount get put in the letter, properly formatted (with the $)?

That sort of thing.

students/EricAdams/session06/test_mailroom.py
```diff
@@ -0,0 +1,221 @@
...
+    :data - a text file
+            in the current directory, (test_send_thank_you_list_inputs.txt),
+            is used as input data.
+            The value that the user would input is simulated
+            by making a one entry into the file per user input.
+            Example: for a user input of 'list', enter 'list'
+            on the first line, (no quotes).
+            Enter 'menu' on the second line for the second user
+            input, (this also forces a return).
+    Modified from a solution to the 'input' issue on stackoverflow.
```
PythonCHB, line 112: nice job! 

This is called "monkey patching" and is a good way to test this kind of thing.

However, it's a bit sensitive to errors. If there is an error between re-assigning sys.stdin and putting it back, then yoursystem will be left in an ustable state.

What you are doing is "mocking" -- and the mock module gives you a way to do it that is more robust -- essentially handling the "make sure it's put back the way it was" part for you -- plus other features.

we'll get into that more when we get into advanced testing techniques in the next quarter.

students/EricAdams/session06/test_mailroom.py
```diff
@@ -0,0 +1,221 @@
...
+    with open("test_send_thank_you_donor_existing_amount_no_write.txt") \
+            as f:
+        donor = f.readline()
+        donor = donor.strip()
+        amount = f.readline()
+    assert mailroom.donor_db[donor][-1] == float(amount)
+    assert result is None
+
+
+def test_send_thank_you_print_letter_no_write(capsys):
```
PythonCHB, line 180: does this need to be tested?

in a way, you are testing that print() works :-)

If you have tests that test that the letter is correctly generated, then maybe no need to test that it gets printed to the console correctly.

students/EricAdams/session06/test_send_thank_you_write_letter_to_file.txt
```diff
@@ -0,0 +1,3 @@
+Jeff Bezos
```
PythonCHB, line 1: are these being generated by the tests? if so, you don't want to put them in git.

students/EricAdams/session07/html_render.py
```diff
@@ -0,0 +1,46 @@
...
+            try:
+                # all_content += each
+                text += each
+            except TypeError:
+                each.render(file_obj)
+        html_tag = '</' + self.tag + '>'
+        text += html_tag
+        self.write_to_file(file_obj, text)
+
+    def write_to_file(self, file_obj, stuff_to_print):
```
PythonCHB, line 32: why a separate method for this?

students/EricAdams/session07/html_render.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def append(self, content):
+        self.content.append(content)
+
+    def render(self, file_obj, indent=''):
+        # all_content = ('<' + self.tag + '>')
+        html_tag = ('<' + self.tag + '>')
+        self.write_to_file(file_obj, html_tag)
+        text = ""
+        for each in self.content:
```
PythonCHB, line 22: this is going to be putting things out of order if you have text and elements in there -- have you got a failed test for this yet?

students/EricAdams/session07/test_html_render.py
```diff
@@ -0,0 +1,165 @@
...
+    more_stuff = 'eggs, eggs, eggs'
+    el_object.append(more_stuff)
+    contents = render_element(el_object)
+    assert contents.startswith('<html>')
+    assert contents.endswith('</html>')
+    assert my_stuff in contents
+    assert more_stuff in contents
+
+
+def test_render_non_strings():
```
PythonCHB, line 138: I see you have my updated tests now -- so this is failing, yes?

I"ll merge and look forward to seeing where this goes.

PythonCHB (approved):
really nice job!

But do be careful about a few things:

1) make sure you test what *matters* about a function's result -- it's easiest (particularly if you wrote the code first) to simply match results, but your system will be more flexible if you test for the parts that matter, and won't change.

2) Ideally, your tests should be as isolated as possible. So if you, for instance, need to test that the correct letter is generated from a donor object, then create a donor object in the test, and pass that in, rather than using pulling it from the donor_db -- that way, the donor_db could be broken, and the individual tests will pass.

This will start to make more and more sense as we do more testing -- and particularly when we do TDD and write the tests along with the code.

PR #124, opened 2017-11-14
Session 06 Mailroom Test code

HiroTestStudent:
Hi Maria,

I just commited and pushed my mailroom5.py. As you suggested on Slack, I took [1] out of donor_dict[d][1]. Now it is working again. 

I still need to refactor the code to make it more testable. 

Thanks,
Hiro


students/hirot_directory/session05/mailroom5.py
```diff
@@ -29,7 +33,7 @@ def print_list():
                                            msg4="Average Gift"))
     for d in donor_dict:
         t = sum(donor_dict[d])
```
PythonCHB, line 35: when looping thourhg a dict, if you find you need both the key and the value, you can do:

```
for donor, donations in donr_dict.items():
    t = sum(donations)
    n = len(donations)
   ...
```

students/hirot_directory/session05/mailroom5.py
```diff
@@ -55,54 +59,30 @@ def quit():
             print("Good bye...")
             sys.exit()
         else:
-            return mainloop()
+          return True # refactoring
```
PythonCHB, line 62: good idea....

students/hirot_directory/session05/mailroom5.py
```diff
@@ -55,54 +59,30 @@ def quit():
...
 
-    except ValueError as e:
-        print("Error happened. Error is ",type(e),"Please type 1, 2 or 3")
     except KeyError as k:
         print("Error happened. Error is ",type(k),"Please type 1, 2 or 3")
```
PythonCHB, line 65: you don't generally want to report the actual error to the user, simple the fact that they typed an invalid input.

students/hirot_directory/session05/mailroom5.py
```diff
@@ -116,30 +96,28 @@ def send_letters():
...
             dict_answer[answer]()
 
-        except ValueError as e:
-            print("Error happened. Error is ", type(e), "Please type 1, 2 or 3")
         except KeyError as k:
-            print("Error happened. Error is ", type(k), "Please type 1, 2 or 3")
-        # finally:
-        #   print("please type 1, 2, or 3")
-
+            print("Sorry...Program crashed....Error type is....", type(k), "Please type 1, 2 or 3")
```
PythonCHB, line 120: did the program crash? or did you catch the error? it can only be one or the other....

students/hirot_directory/session05/test_mailroom.py
```diff
@@ -0,0 +1,18 @@
...
+"""
+
+import pytest
+import mailroom5 as mr
+
+
+def test_donor_dict():
+    """ Test to make sure we make put everything in a dictionary. """
+    result = mr.donor_dict
+    assert 'Paul Allen' in result.keys()
```
PythonCHB, line 17: This is just `assert "Paul Allen" in result`

students/hirot_directory/session06/test_mailroom.py
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env python3
```
PythonCHB, line 1: need a bit more here.... :-)

students/hirot_directory/session06/mailroom6_3.py
```diff
@@ -0,0 +1,143 @@
+#!/usr/bin/env python3
```
PythonCHB, line 1: do let us know which version you want reviewed, when there is more than one in a PR.

students/hirot_directory/session06/mailroom6_3.py
```diff
@@ -0,0 +1,143 @@
...
+
+"""
+
+import sys
+
+donor_names = ["William Gates III", "Mark Zuckerberg", "Jeff Bezos", "Paul Allen"]
+donations = [[200, 500], [2000, 3000], [4000, 5000], [100, 150]]
+
+donor_dict = { donor_names:donations for (donor_names, donations) in zip(donor_names, donations)}
+
```
PythonCHB, line 15: you can also do with:

`donor_dict = dict(zip(donor_names, donations))`

students/hirot_directory/session06/mailroom6_3.py
```diff
@@ -0,0 +1,143 @@
...
+
+def print_list():
+    # print(donor_dict)
+    print("{msg1: <50}| {msg2: <12}| " \
+
+          "{msg3:<10}| {msg4: <12}".format(msg1="Donor Name",
+
```
PythonCHB, line 22: extra empty lines here ?

students/hirot_directory/session06/mailroom6_3.py
```diff
@@ -0,0 +1,143 @@
...
+        t = sum(donor_dict[d])
+        n = len(donor_dict[d])
+        a = t / n
+
+        print("{d: <50} ${t: 12.2f}{n: 12d}{a: 14.2f}".format(d=d, t=t, n=n, a=a))
+
+
+def make_donation():
+    donor_name = str(input("Full Name: "))
+    new_donation = int(input("How much do you want to donate?: "))
```
PythonCHB, line 38: what if they don't type a number? 

you probably want some error-catching here.

students/hirot_directory/session06/mailroom6_3.py
```diff
@@ -0,0 +1,143 @@
...
+def quit():
+    print("quit function is called")
+
+    try:
+        selection = (input("Are you sure to quit? (Y or N) To return to main menu, please select 'N'.: ")).upper()
+        if (selection == 'Y'):
+            print("Good bye...")
+            sys.exit()
+        else:
+            return mainloop() # refactoring candidate
```
PythonCHB, line 56: yes -- you don't want to be calling mainloop multiple times -- this needs to simply return, maybe with a "success" value of some sort.

students/hirot_directory/session06/mailroom6_3.py
```diff
@@ -0,0 +1,143 @@
...
+            print("Error happened. Error is ", type(e) , "Please type 1, 2 or 3")
+        except KeyError as k:
+            print("Error happened. Error is ", type(k) , "Please type 1, 2 or 3")
+
+
+
+
+
+def create_report():
+    print("{msg1: <50}| {msg2: <12}| " \
```
PythonCHB, line 87: If you are going to use generic names like msg1 -- might as well simply use positional arguments:

```
print("{msg1: <50}| {msg2: <12}|...

.format("Donor Name", "Total Given", ... )
```

PythonCHB: no need for the `\` -- when inside a parentheses, you don't need it.

students/hirot_directory/session06/mailroom6_3.py
```diff
@@ -0,0 +1,143 @@
...
+        try:
+            answer = int(input("Select from one of these options:\n"
+                               "(1) Send a Thank you\n"
+                               "(2) Create a Report\n"
+                               "(3) Send letters to everyone\n"
+                               "(4) quit\n"
+                               ))
+            dict_answer[answer]()
+
+        except ValueError as e:
```
PythonCHB, line 133: nicely done!

students/hirot_directory/session06/mailroom6_3.py
```diff
@@ -0,0 +1,143 @@
...
+            text = "Dear Mr./Ms. {},\n\n    Thank you for your kind donation of USD ${}.\n    It will be put to very good use.\n\n          Sincerely,\n            -The Team".format(n, d).encode()
+            f.write(text)
+            f.close()
+
+    except IOError as i:
+        print("Error code is", i)
+    print("letter drafts are in the folder for your review. Please review before mailing out.")
+
+
+dict_answer = {1: thank_you_loop, 2: create_report, 3: send_letters, 4: quit}
```
PythonCHB, line 120: might want to move this inside the mainloop function -- it will still only be defined once, as you call mainloop only once (or should).

PythonCHB (approved):
I look forward to seeing it tested :-)

But make sure to keep up with thenew stuff!

I think at this point, better to be doing testing with the html render exercise than more time on mailroom, if you don't have time for both.

And we'll be extending mailroom to be OO too :-)

PR #126, opened 2017-11-20
Steps 1-3 on html assignment

Eowyn42:
Finished and tested (98% coverage!) steps 1 - 3 of the html rendering assignment. AFAIK this is as far as I'm supposed to get at this stage -- correct me if that's wrong. I have the output looking just like the sample for 1,2 and almost like it for 3 -- the indents on the inner content are hard to do as they appear to be different than the other indents (self.indent) that I have access to?
<title>PythonClass = Revision 1087:</title>
should be
<title>  PythonClass = Revision 1087:  </title>

### DIFF
--- a/students/eowyn/session07/html-rendering/test_html_render.py
+++ b/students/eowyn/session07/html-rendering/test_html_render.py
@@ -1,0 +1,81 @@
+import html_render
+from html_render import Element, Body, P, Html
+
+def test_new_element():
+    el_obj = Element
+    el_obj2 = Element('content')
+
+
+def test_add_content():
+    el_object = Element('content')
+    el_object = Element()
+    assert el_object.content == []
+
+
+def test_adding_empty_string():
+    el_object = Element('')
+    assert el_object .content == ['']
+
+
+def test_append_string():
+    el_object = Element('spam, spam, eggs')
+    el_object.append(' and spam')
+    assert el_object.content == ['spam, spam, eggs', ' and spam']
+
+
+def test_tag_exists():
+    assert Element.tag == 'html'
+    el_object = Element('spam, spam, spam')
+    assert el_object.tag == 'html'
+
+
+def test_indent_exists():
+    # could alternately test that it is any number of spaces
+    assert Element.indent == 4 * ' '
+
+
+def test_render():
+    my_stuff = 'spam, spam, spam'
+    more_stuff = '\neggs, eggs, eggs'
+    el_object = Element(my_stuff)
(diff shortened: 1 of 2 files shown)


PythonCHB:
yes, steps 1-3 is good -- we'll pick up with step 4 in the next class.

students/eowyn/session07/html-rendering/test_html_render.py
```diff
@@ -0,0 +1,105 @@
...
+def test_render_body():
+    my_stuff = 'spam, spam, spam'
+    more_stuff = '\neggs, eggs, eggs'
+    el_object = Body(my_stuff)
+    el_object.append(more_stuff)
+    with open('test1', 'w') as out_file:
+        el_object.render(out_file)
+    with open('test1', 'r') as in_file:
+        contents = in_file.read()
+    assert contents.startswith('\n<body>')
```
PythonCHB, line 70: you might want to .strip() the contents, and then check startswith -- as whitespace is meaningless to html.

students/eowyn/session07/html-rendering/test_html_render.py
```diff
@@ -0,0 +1,105 @@
...
+
+def test_html_tag():
+    assert Html().tag == 'html'
+
+def test_render_body():
+    my_stuff = 'spam, spam, spam'
+    more_stuff = '\neggs, eggs, eggs'
+    el_object = Body(my_stuff)
+    el_object.append(more_stuff)
+    with open('test1', 'w') as out_file:
```
PythonCHB, line 66: see my code for how you can put this writing the reading a file in a utility function for your tests.

students/eowyn/session07/html-rendering/test_html_render.py
```diff
@@ -0,0 +1,105 @@
...
+    assert more_stuff in contents
+
+def test_render_non_strings():
+    my_stuff = 'any string I like'
+    el_object = Element(Body(my_stuff))
+    with open('test3', 'w') as out_file:
+        el_object.render(out_file)
+    with open('test3', 'r') as in_file:
+        contents = in_file.read()
+    assert my_stuff in contents
```
PythonCHB, line 82: did the element get rendered??

(see my tests for this, too :-) )

PythonCHB (approved):
looking good! 

you've got the basic logic for indenting down, but could maybe clean it up a bit.

careful here -- having near 100% test coverage is great, but make sure you are testing for all the key things that should happen...

PR #127, opened 2017-11-21
htm_render.py up to step 3

ea8606:


ea8606:
Not sure if the way I write the tags to a file first , process the text and then write the text to the same file is the best way  to go.

PythonCHB:
does everything get rendered in the right order?

but I'll take a look.

students/EricAdams/session07/html_render.py
```diff
@@ -3,7 +3,9 @@
 class Element():
 
     tag = 'html'
-    indent = '  '
+    indent = ''
+    # keep track of the depth of recursion
+    depth = 0
```
PythonCHB, line 8: ideally you don't need to do this at all, though it's a fine approach.

But you definitely don't want depth as a class attribute -- class attributes are for when all the intances should share the same value -- certainly not the case here!

students/EricAdams/session07/html_render.py
```diff
@@ -15,18 +17,21 @@ def append(self, content):
         self.content.append(content)
 
     def render(self, file_obj, indent=''):
-        # all_content = ('<' + self.tag + '>')
-        html_tag = ('<' + self.tag + '>')
+        # what recursion level are we going in
+        Element.depth += 1
```
PythonCHB, line 21: problem here -- every single Element will now have a larger recursion depth!

If this is working, then it's because you are changing it constantly to keep in sync.

But better to simply pass it through when you render.

That's what the indent parameter was for -- to pas in the current level of indentation -- you could pass in a "depth" instead. but better to pass it through that store it in the Element class.

students/EricAdams/session07/html_render.py
```diff
@@ -15,18 +17,21 @@ def append(self, content):
...
-        html_tag = '</' + self.tag + '>'
-        text += html_tag
+        # what recursion level are at coming out
+        Element.depth -= 1
+        html_tag = '\n' + (self.indent * (Element.depth)) + \
+            '</' + self.tag + '>\n'
+        text = self.indent * (Element.depth) + text + html_tag
         self.write_to_file(file_obj, text)
 
     def write_to_file(self, file_obj, stuff_to_print):
```
PythonCHB, line 37: no reason to have a method for this --it's one line of code, and doesn't use self...

students/EricAdams/session07/html_render.py
```diff
@@ -35,12 +40,45 @@ def write_to_file(self, file_obj, stuff_to_print):
 
 class Body(Element):
     tag = 'body'
+    indent = '    '
```
PythonCHB, line 43: you don't want to redefine this here. -- unless you want each type of Element to have a different indent level -- probably not. If you define indent only in the base class, then all subclasses will use the same one.

then you can change it in the base class, and the change is seen everywhere.

students/EricAdams/session07/html_render.py
```diff
@@ -35,12 +40,45 @@ def write_to_file(self, file_obj, stuff_to_print):
...
+    tag = 'head'
+    indent = '    '
+
+
+class OnelineTag(Element):
+    # override Element.render to write all on one line
+    def render(self, file_obj, indent=''):
+        # what recursion level are we going in
+        Element.depth += 1
+        html_tag = (self.indent * (Element.depth - 1) + '<' + self.tag + '>')
```
PythonCHB, line 66: since you have to subtract one -- maybe just add one after using it instead.

students/EricAdams/session07/html_render.py
```diff
@@ -35,12 +40,45 @@ def write_to_file(self, file_obj, stuff_to_print):
...
+        file_obj.write(html_tag)
+        text = ""
+        for each in self.content:
+            try:
+                text += each
```
PythonCHB, line 71: this is where things can get out of order -- better to render either the text or the sub-element to the file in the same loop.

students/EricAdams/session07/test_html_render.py
```diff
@@ -155,11 +155,77 @@ def test_render_non_strings():
...
+
+    print(contents)  # so we can see what's going on if a test fails
+
+    # so what should the results be?
+    # the html tag is the outer tag, so the contents should start
+    # and end with that.
+    assert contents.startswith('<html>')
+    assert contents.endswith('</html>')
+
+    # the body tags had better be there too
```
PythonCHB, line 184: looking good!

though you may want to set the number of spaces somewhere in your test, and then use it. maybe even:

the_indent = Element.indent

students/EricAdams/session07/test_html_render.py
```diff
@@ -155,11 +155,77 @@ def test_render_non_strings():
...
+
+    # now lets get pretty specific:
+    # the opening tag should come before the ending tag
+    assert contents.index('    <body>') < contents.index('    </body>')
+    # the opening tag should come before the <p> tag
+    assert contents.index('    <body>') < contents.index('<p>')
+    # opening tag before the ending tag
+    assert contents.index('        <p>') < contents.index('        </p>')
+    assert contents.index('<p>') < contents.index('any string I like')
+
```
PythonCHB, line 203: and now the REAL test.

change Element.indent, and see if the indentation is all still correct!

students/EricAdams/session07/html_render.py
PythonCHB: though it is working, yes?

but did you test arbitrary mixing of sub-elements and text?

PythonCHB (approved):
looking good!

PR #128, opened 2017-11-25
Html renderer

guillaumethomas:
Hello

Thank you for looking at it!
I may come come on Sunday, I have some questions on edge cases regarding kwargs and class design.

Thank you


PythonCHB:
see you there.

students/guillaume/session07/test_html_render.py
```diff
@@ -0,0 +1,65 @@
...
+        el_object.render(out_file)
+    with open('test1', 'r') as in_file:
+        contents = in_file.read()
+    assert contents.startswith('<html>')
+    assert contents.endswith('</html>')
+    assert my_stuff in contents
+    assert more_stuff in contents
+
+
+def test_body_tag():
```
PythonCHB, line 49: I'll talk about this in class, but you don't want to test too much implementation detail --and having the tag attribute is an implementation detail.

students/guillaume/session07/html_render.py
```diff
@@ -0,0 +1,129 @@
...
+
+
+class Hr(SelfClosingTag):
+    tag = 'hr'
+    indent = 4 * ' '
+
+
+class Br(SelfClosingTag):
+    tag = 'br'
+    indent = 4 * ' '
```
PythonCHB, line 128: I think we talked about this, but you don't want to set the indent for each type of element -- they should all share the same indentation style -- so it should be a class attribute of the base class.

students/guillaume/session07/html_render.py
```diff
@@ -0,0 +1,129 @@
...
+
+    def __init__(self, level, content=None):
+        OneLineTag.__init__(self, content)
+        self.tag = '{}{}'.format(self.tag, str(level))
+
+
+class SelfClosingTag(Element):
+
+    def render(self, file_obj, ind=""):
+        att = ', '.join(['{}="{}"'.format(arg, self.kwargs[arg])
```
PythonCHB, line 60: This looks a lot like the code in the Element render method -- maybe factor that out so as not to repeat code.

students/guillaume/session07/html_render.py
```diff
@@ -0,0 +1,129 @@
...
+    indent = ''
+
+    def __init__(self, content=None, **kwargs):
+        self.kwargs = kwargs
+        if content is None:
+            self.content = []
+        else:
+            self.content = [content]
+
+    def append(self, content=None, **kwargs):
```
PythonCHB, line 14: when you are adding more content to a tag -- is it appropriate to add attributes then? 

Probably not.

If you want to change the attributes of an existing element, you can do that by simply adding to the attribute dict:

the_element.kwargs["an_attribute"] = the_value_of_the_attribute

PythonCHB: And this kind of makes it clear why "kwargs" is not a great name for the attribute...

PythonCHB (approved):
looking good -- but do write some more tests!

PR #129, opened 2017-11-27
Cleaned up html renderer with expanded testing

Eowyn42:
This already had one review. The logic hasn't changed much but I'm testing quite a bit more; a quick review of the tests would be helpful. Thank you.

Revisions: 
- Added render_element function to test_html_renderer to simplify the test framework

- Added testing for indendation, imbedded elements, one line elements

- Expanded tests of body, paragraph, head objects beyond the tag

- Added contents.strip() to all tests to negate testing white space

- Cleaned up the renderer code to more accurately and clearly insert the right carriage returns. It does mostly the same thing as before but in a better-tested, more robust way.

- Cleaned up the renderer code to have the LOCAL current_indent be more clearly separated from the CLASS extra_indent ; the class variable describes how the indent changes as you go deeper into the nest essentially

- Adjusted the one line render function to NOT be recursive. This was a judgment call, but I decided that the idea of having it handle "simple tags" should be a strict one. I decided not to throw an error if file object was passed in, I'm just printing the object's alphanumeric representation. 

- Added some commentary


students/eowyn/session07/html-rendering/html_render.py
```diff
@@ -1,35 +1,32 @@
-
-
 class Element():
+    # Pretty-print HTML files with increasing indent levels
+    # for each depth of tag.
     tag = 'html'
-    indent = 4 * ' '
+    extra_indent = 4 * ' '
```
PythonCHB, line 5: is this "extra"? or the basic amount.

If you want to be more explicit, maybe "indent_size" or something...

though if "extra' means something to you -- go for it!

students/eowyn/session07/html-rendering/html_render.py
```diff
@@ -1,35 +1,32 @@
...
-        if self.tag == 'html':
-            out_file.write('<!DOCTYPE html>')
-        out_file.write('\n')
-        out_file.write(ind + '<' + self.tag + '>')
+    def render(self, out_file, current_ind=""):
+        # Pretty-print content and tags to file.
+        # For objects, pretty-print all the content and tags
+        # they contain to the file. Nest tags with indentation.
+
+        out_file.write('\n' + current_ind + '<' + self.tag + '>')
```
PythonCHB, line 22: this is going to put a newline in before every tag -- including the first one -- maybe not quite the right place?

PythonCHB (approved):
looking great -- love the additional tests...

PR #130, opened 2017-11-27
Session 06

kathrynegan:
I did some housekeeping so it looks a little frightening. I only need a pair of eyes on html_render.py and test_html_render.py, though. Thanks!

And I meant Session 7... this quarter is going by too fast!


PythonCHB:
yes, it is going fast!

You have two conflicting files:

```
students/kegan/session03/mailroom.py
students/kegan/session06/argparse_demo.py
```
that means the version in your repo is different than the version in the class repo, and in a way that git can't figure out how to resolve.

To clean this up:

do a `git pull upstream master` on your personal machine.

this will merge what's in the class repo with yours.

if you look at the messages, there will probably be "conflicts" in those two files.

open the files, and look for things like:

>>>>>>>>>>

and
<<<<<<<<<<

what git does when the same lines are changed in both versions, is put BOTH in the merged file, separted by those symbols. 

you then edit the file to remove th markers and keep the parts you want.

after doing that, you can commit the changes:

git commit -a

and then push to your gitHub repo:

git push

and the PR should be automatically updated, and hopefully not have the conflicted files anymore.

students/kegan/session07/html_render.py
```diff
@@ -0,0 +1,186 @@
...
+
+"""
+
+
+class Element:
+    """ Stores behavior and data for an Element of HTML. """
+    indent = '    '
+
+    def __init__(self, content=None, **kwargs):
+        """ Initializes an Element with content, if given,
```
PythonCHB, line 12: nice docstring!

students/kegan/session07/html_render.py
```diff
@@ -0,0 +1,186 @@
...
+            content (str or Element) : text or a nested Element
+        """
+        self.content.append(content)
+
+    def render(self, file_out, ind=0):
+        """ Renders the contents of this Element into HTML
+        format and pretty prints to file using given file handler.
+        Args:
+            file_out (TextIOWrapper) : file handler
+            ind (str) : level of indentation for this Element
```
PythonCHB, line 36: this says ind is a string, but the default is zero...

and looking later, it is the number of levels of indentation, not the actual indentation, which is a fine way to do it, but make sure to keep your docstrings up to date.

students/kegan/session07/html_render.py
```diff
@@ -0,0 +1,186 @@
...
+    """ Handles contentless HTML elements. """
+
+    def render(self, file_out, ind=0):
+        """ Renders this contentless Element into HTML
+        and writes to file using given file handler.
+        Args:
+            file_out (TextIOWrapper) : file handler
+            ind (str) : level of in
+        """
+        file_out.write(self.get_tag(ind, 'selfclosing'))
```
PythonCHB, line 99: given that pretty much all the render method does is the tag itself, the code for it might as well be here.

students/kegan/session07/html_render.py
```diff
@@ -0,0 +1,186 @@
...
+        file_out.write('\n')
+        for c in self.content:
+            try:
+                c.render(file_out, ind + 1)
+            except AttributeError:
+                file_out.write('{}{}\n'.format((ind + 1) * self.indent, c))
+        file_out.write(self.get_tag(ind, style='close'))
+        file_out.write('\n')
+
+    def get_tag(self, ind, style='open'):
```
PythonCHB, line 48: this is pretty slick.

I"m on the fence about whether this is worth it, rather than haveing two methods:

get_open_tag and get_close_tag.

or, what I did, was have the method make both tags at once, and return them -- after all you always need both. except the self-closing tag, which could be an overridden method in the self-closing-tag class.

that last point is key -- the base class shouldn't have to know about something that only some of the subclasses need.

students/kegan/session07/html_render.py
```diff
@@ -0,0 +1,186 @@
...
+        Returns:
+            tag : formatted tag
+        """
+        substrings = {
+            'indent': ind * self.indent,
+            'tag': self.tag,
+            'attributes':
+                '' if style == 'close' else ''.join([
+                    ' {}="{}"'.format(key, self.attrs[key])
+                    for key in sorted(self.attrs)]),
```
PythonCHB, line 64: why sorted? I don't think order matters to attributes.

students/kegan/session07/html_render.py
```diff
@@ -0,0 +1,186 @@
...
+    """ HTML hyperlink element. """
+    tag = 'a'
+
+    def __init__(self, link, content):
+        """ Maps link to content as an HTML attribute of an Element.
+        Args:
+            link (str) : hyperlink
+            content (str) : content to map hyperlink to
+        """
+        Element.__init__(self, content, href=link)
```
PythonCHB, line 128: nice! pretty slick, isn't it?

students/kegan/session07/html_render.py
```diff
@@ -0,0 +1,186 @@
...
+        """
+        file_out.write('<!DOCTYPE {}>\n'.format(self.tag))
+        Element.render(self, file_out, ind=1)
+
+
+class A(OneLineTag):
+    """ HTML hyperlink element. """
+    tag = 'a'
+
+    def __init__(self, link, content):
```
PythonCHB, line 122: I can't remember how I wrote this up, but it turns out that anchor tags can have other attributes:

https://www.w3schools.com/tags/tag_a.asp

so you should probably have a **kwargs here, too.

students/kegan/session07/html_render.py
```diff
@@ -0,0 +1,186 @@
...
+            link (str) : hyperlink
+            content (str) : content to map hyperlink to
+        """
+        Element.__init__(self, content, href=link)
+
+
+class H(OneLineTag):
+    """ HTML header element. """
+
+    def __init__(self, level, content):
```
PythonCHB, line 134: headers can also take attributes -- I need to go and look at what I wrote in the Exercise !

students/kegan/session07/test_html_render.py
```diff
@@ -0,0 +1,208 @@
+"""
+Kathryn Egan
+
+Suite of tests for components of html_render.py
+"""
+from html_render import *
```
PythonCHB, line 6: `import *` is frowned upon -- but this is one place it probably makes sense :-)

students/kegan/session07/test_html_render.py
```diff
@@ -0,0 +1,208 @@
...
+#     html.append(Body('body content'))
+#     html.append(P())
+#     temp = 'html_test.html'
+#     with open(temp, 'w') as f:
+#         html.render(f)
+#     with open(temp, 'r') as f:
+#         assert f.read() == '<html><body>body content</body><p></p></html>'
+
+
+def prepare(elements):
```
PythonCHB, line 36: careful here -- you are testing against a specific result -- makes the tests not very flexible if you change how you want things rendered.

students/kegan/session07/test_html_render.py
```diff
@@ -0,0 +1,208 @@
...
+    Returns:
+        str : elements as rendered HTML
+    """
+    output = '\n'.join([
+        '    ' * (indent + 1) + element for indent, element in elements])
+    output = '<!DOCTYPE html>\n{}\n'.format(output)
+    return output
+
+
+def check_render(html, expected):
```
PythonCHB, line 50: good idea to put this in a function, but careful about doing checks for specific complete renders -- not very flexible.

better to have this function return what was rendered, and do the comparison in the tests.

students/kegan/session07/test_html_render.py
```diff
@@ -0,0 +1,208 @@
...
+    html.append(Title('this is the title'))
+    elements = [
+        (0, '<html>'),
+        (1, '<title>this is the title</title>'),
+        (0, '</html>')]
+    expected = prepare(elements)
+    check_render(html, expected)
+
+
+def test_self_closing_tag():
```
PythonCHB, line 91: ideally, these would be tested by themselves -- does a self closing tag render itself properly? this is almost that, but it's also testing that the html element does the right thing.

PythonCHB (approved):
nice work!

the only key issue is the teting against complete known output -- not very flexible in the face of any changes.

You are testing the system rather than the individual parts -- ideally unit tests test just one little piece at a time. you may test how they all fit together as well, but only after the pieces are tested.

PythonCHB:
All good -- but I won't merge until the conflicts are cleaned up.

kathrynegan:
OK, I think I successfully carried out your instructions. I didn't see
any <<<<<<<<<<
markers in the two files so pushed them as-is.

PR #132, opened 2017-11-28
html_render & test_html_render woking in progress

HiroTestStudent:
Hi Chris, 

Thank you for help in office hours on Sunday. It was very helpful. I worked a little bit on html_render and test_html_render files. I haven't completed all the steps yet (I am in step 5) but please let me git push what I have now.

I am not sure of whether I wrote test file correctly but all tests passed for now. 

Thanks!
Hiro


students/hirot_directory/session07/test_html_render.py
```diff
@@ -0,0 +1,246 @@
...
+
+    print(contents)
+
+    lines = contents.split('\n')
+
+    # assert len(lines) == 3
+    assert lines[0].startswith('  <')
+    assert lines[1].startswith('  ' + Element.ind)
+    assert lines[2].startswith('  <')
+
```
codedragon, line 204: This test passes, but it is only checking one element. When I tried to do more than one element, this is what was produced:
```
  <html>
    <body>
<p>
some text
</p>
</body>
  </html>
```
I will add comments in your render method to help you figure out what is going on here.

students/hirot_directory/session07/test_html_render.py
```diff
@@ -0,0 +1,246 @@
...
+    contents = render_element(el_object, remove=False) # adding remove=False to see the output
+    # make sure extra whitespace at beginning or end doesn't mess things up.
+    contents = contents.strip()
+
+    print(contents)  # so we can see what's going on if a test fails
+
+    assert contents.startswith('<') # Element.ind+'any'
+    assert contents.endswith('>')
+
```
codedragon, line 238: Your one line tag works, but this test isn't testing it properly. Your content could be just 
```
<html>
```
and this would pass. you really want startswith('<title>') and endswith('</title>'). Also, you want to make sure there is only one line of text in the file, so that you know that it really is a one line tag. :)

students/hirot_directory/session07/html_render.py
```diff
@@ -0,0 +1,96 @@
...
+                file_obj.write(each)
+            except TypeError:
+                each.render(file_obj)
+
+        file_obj.write('\n')
+
+        file_obj.write(ind)
+        closing_tag = ('</' + self.tag + '>')
+        file_obj.write(closing_tag)
+
```
codedragon, line 39: Indenting is difficult, so don't beat yourself up too much about this. Note that there are two different ind variables. There is the self.ind variable, and there is the ind that comes in as an optional input variable to render. The self.ind variable is what is used by each element to ensure that it is indented the way it should be. The reason for having an optional ind variable in your render method is so that you can pass in the ind for each element that is being rendered, so that render can keep track of the total amount of indentation for any given line. For example, if you have a body element, and a paragraph element inside the body element, then the paragraph line needs to be indented both the amount of indentation for the body plus the amount of indentation for the paragraph.
So, outside the try/except, you use ind to indent the tag, since the tag will be whatever the previous indentation was already. In the try/except, you are indenting the new element, so you want to use self.ind + ind, so that it is now indented more, by however much this new element should be indented. If there is no content (if your try fails), then you instead you should add the indentation of the new element to the previous indentation, iow, ```each.render(file_obj, ind + self.indent)```.  Now you should not have to write(ind) twice. 
Now, however, you have a new problem, because of recursion. Your indent for the before tag and after tag will not be the same, because you are writing some indent before the write fails (same problem Chris had in class). You will need a similar solution to what Chris came up with.

codedragon (approved):

PR #133, opened 2017-11-28
html_render step 3... <p> . .. can't git rid of the extra indents

ea8606:


PythonCHB:
still having problems with the extra indents in <p> elements?

students/kegan/session07/test_html_render.py
```diff
@@ -0,0 +1,208 @@
...
+#     html.append(Body('body content'))
+#     html.append(P())
+#     temp = 'html_test.html'
+#     with open(temp, 'w') as f:
+#         html.render(f)
+#     with open(temp, 'r') as f:
+#         assert f.read() == '<html><body>body content</body><p></p></html>'
+
+
+def prepare(elements):
```
PythonCHB, line 36: this makes me a little nervous -- kinda fixed results expected -- but great instinct on automating the testing process. And I guess if you do change anything, you can update this one function :-)

students/kegan/session07/html_render.py
```diff
@@ -0,0 +1,186 @@
...
+        """
+        file_out.write('<!DOCTYPE {}>\n'.format(self.tag))
+        Element.render(self, file_out, ind=1)
+
+
+class A(OneLineTag):
+    """ HTML hyperlink element. """
+    tag = 'a'
+
+    def __init__(self, link, content):
```
PythonCHB, line 122: I actually looked it up -- anchor tags can take other attributes as well. So this should take attribures as kwargs and add the href one to them.

But the right approach, for sure.

students/kegan/session07/html_render.py
```diff
@@ -0,0 +1,186 @@
...
+    """ Handles contentless HTML elements. """
+
+    def render(self, file_out, ind=0):
+        """ Renders this contentless Element into HTML
+        and writes to file using given file handler.
+        Args:
+            file_out (TextIOWrapper) : file handler
+            ind (str) : level of in
+        """
+        file_out.write(self.get_tag(ind, 'selfclosing'))
```
PythonCHB, line 99: the idea of subclassing is the the subclass-specific stuff in the subclass, rather than the Element get_tag method needing to know about self closing tags...

students/kegan/session07/html_render.py
```diff
@@ -0,0 +1,186 @@
...
+
+    def render(self, file_out, ind=0):
+        """ Renders the contents of this Element into HTML format
+        on one line and writes to file using given file handler.
+        Args:
+            file_out (TextIOWrapper) : file handler
+            ind (str) : level of indentation for this Element
+        """
+        file_out.write(self.get_tag(ind))
+        for c in self.content:
```
PythonCHB, line 83: should one-line tags support multiple pieces of content?

honestly, I'm not totally sure, there are no rules, as whether it's on one line or not makes no difference to html itself.

students/kegan/session07/html_render.py
```diff
@@ -0,0 +1,186 @@
...
+        """ Generates an HTML tag according to given style
+        with given level of indent. Style can be 'open',
+        'close', or 'selfclosing' (defaults to 'open').
+        Args:
+            ind (str) : level of indent
+            style (str) : style to use
+        Returns:
+            tag : formatted tag
+        """
+        substrings = {
```
PythonCHB, line 58: I'm not sure putting it all in a dict, just so you can pass it in to format it worth it.

though it is kind of elegant.

PythonCHB (approved):
let me know if you want anything else reviewed.

Really nice job on the html_render!

students/kegan/session07/html_render.py
ea8606: everything on the html_render is ok.   Learned lots from your solution

PR #135, opened 2017-11-29
implemenation of circle and sparse array; added OO approach to mailroom

mattmaeda:


PythonCHB:
That was fast!

students/mattmaeda/session08/circle.py
```diff
@@ -0,0 +1,138 @@
...
+        :param diameter: the circle's diameter
+        :type: int or float
+        """
+        if Circle.is_valid_value(diameter):
+            return Circle(diameter/2)
+        raise ValueError("invalid diameter")
+
+
+    @staticmethod
+    def is_valid_value(value):
```
PythonCHB, line 31: you don't usually do a lot of checking for type in python.

If you want to assure that you have a proper value for, example, the radius, do a:

radius = float(radius)

and it'll automagically give you a ValueError.

And if you REALLY need to know it's a number, than you can do:

isinstance(x, Number)

See:

https://docs.python.org/3/library/numbers.html#module-numbers

But again, I'd mostly just stick with duck typing -- let people set radius to anything they want, and it'll fail when they try to use it.

students/mattmaeda/session08/circle.py
```diff
@@ -0,0 +1,138 @@
...
+    def __int__(self):
+        return self._radius
+
+
+    def __float__(self):
+        return self._radius
+
+
+    @property
+    def radius(self):
```
PythonCHB, line 100: with a getter and no setter, it can't be changed.

and if you have a property for something, you generally want to use THAT to reference the value everywhere else.

The rule of thumb for "pythonicity" is to use ordinary attributes unless you really need to do something special when they are set or not.

students/mattmaeda/session08/circle.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+    @property
+    def area(self):
+        """ area getter """
+        return pi * self._radius**2
+
+
+    @radius.setter
+    def radius(self, radius_value):
```
PythonCHB, line 118: ahh -- there is a setter. You REALLY want to put this right below the getter, so it's clear that it's there!

and I see that the point of the property it to type-check it when set. OK, if you really need to do that, but again, I don't recommend it.

students/mattmaeda/session08/circle.py
```diff
@@ -0,0 +1,138 @@
...
+        """ radius setter
+        :param radius_value: radius value
+        :type: int or float
+        """
+        if Circle.is_valid_value(radius_value):
+            self._radius = radius_value
+        else:
+            raise ValueError("invalid radius")
```
PythonCHB, line 126: if you are going to do this -- it should probably be a TypeError.

ValueError means the type is OK, but the Value of that type is wrong. IN this case, all you checked was the type.

students/mattmaeda/session08/circle.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+    @diameter.setter
+    def diameter(self, diameter_value):
+        """ diameter setter
+        :param diameter_value: radius value
+        :type: int
+        """
+        if Circle.is_valid_value(diameter_value):
+            self.radius = diameter_value / 2
```
PythonCHB, line 136: notice that this would raise an error when you tried to compute the radius anyway if an oddball type was passed in.

students/mattmaeda/session08/sparse_array.py
```diff
@@ -0,0 +1,74 @@
+#!/usr/bin/env python3
+"""
+Sparse array implementation
+"""
+
+class SparseArray(object):
+    def __init__(self, number_array):
+        self.__length = len(number_array)
+        self.__array = {}
```
PythonCHB, line 9: you want to use single underscores to indicate a "private" variable.

PythonCHB (commented):
I'll finish up later.

students/mattmaeda/session08/test_circle.py
```diff
@@ -0,0 +1,157 @@
...
+    c = Circle(2)
+    assert repr(c) == "Circle(2)"
+
+
+def test_circle_str(capfd):
+    c = Circle(4)
+    print(c)
+    out, err = capfd.readouterr()
+    assert out == "Circle with radius: 4\n"
+
```
codedragon, line 73: This is cool. I don't use pytest in my regular work, so was unaware of this. It is a little strange, not use to seeing input variables for tests unless there is a decorator, but kind of cute.

students/mattmaeda/session09/circle.py
```diff
@@ -0,0 +1,138 @@
...
+    @classmethod
+    def from_diameter(cls, diameter):
+        """
+        :param diameter: the circle's diameter
+        :type: int or float
+        """
+        if Circle.is_valid_value(diameter):
+            return Circle(diameter/2)
+        raise ValueError("invalid diameter")
+
```
codedragon, line 28: A little strange raising ValueError for both value and type, but not terrible. I guess I would agree with Chris, I'd be more inclined to just try it, let type error be raised on its own, and just do the check for greater than zero for the ValueError.

students/mattmaeda/session09/mailroom.py
```diff
@@ -0,0 +1,219 @@
...
+
+        for donor_obj in self.donors:
+            if donor_obj.name == donor_name:
+                donor = donor_obj
+                break
+
+        if donor is None:
+            donor = Donor(donor_name)
+            self.donors.append(donor)
+
```
codedragon, line 124: Good place for an else. Instead of initializing donor with None, have an else in your for loop that does what is in your "if donor is None" block.  Else in a for loop means "else if you did not break out early". ;-) Alternatively, since you aren't doing anything with donor after the for loop besides returning it, you could return donor_obj immediately when it is found, and then anything that runs after the for loop is by default not found in the for loop, so you can initialize donor there.

students/mattmaeda/session09/mailroom.py
```diff
@@ -0,0 +1,219 @@
...
+if __name__ == "__main__":
+    DONORS = {
+        "George Washington": [1],
+        "John Adams": [3],
+        "Thomas Jefferson": [3],
+        "John Quincy Adams": [2],
+        "James Madison": [1]
+    }
+
```
codedragon, line 191: What is this DONORS dictionary for? I don't see it reference anywhere, other than the initialization.

students/mattmaeda/session09/mailroom.py
```diff
@@ -0,0 +1,219 @@
...
+                donation_amount = input("Enter a donation amount: ")
+                print(mailroom_obj.add_donation(name_of_donor,
+                                                int(donation_amount)))
+            elif choice == 2:
+                print(mailroom_obj.report())
+            else:
+                QUIT = True
+
+        except ValueError:
+            print("Invalid choice '{}'".format(choice))
```
codedragon, line 219: This is quite a bit for a "if __name__ == "__main__"" block. I would consider creating a function for user interaction, and just call that function from here.

students/mattmaeda/session09/test_mailroom.py
```diff
@@ -0,0 +1,154 @@
...
+    for donor, amount in INITIAL_DONORS.items():
+        mailroom.add_donation(donor, amount)
+
+    mailroom.add_donation("Monty Python", 10)
+    sorted_donors = [donor_obj.name for donor_obj in
+                     mailroom.sort_donors_by_donation()]
+
+    assert ["Monty Python", "John Adams", "Thomas Jefferson",
+            "John Quincy Adams", "George Washington",
+            "James Madison"] == sorted_donors
```
codedragon, line 154: I would iterate through your INITIAL_DONORS dictionary to create the list for this assertion. That way if you change your initial conditions, you don't have to change your tests.

codedragon:
I finished up for Chris. If you have any questions, let me know. It looked like the circle and circle tests were just copied from one directory to another, with no changes. Any reason for that?
Good job with the mailroom exercise and tests, looks great!

codedragon:
Ack, meant to test your sparse array code, but didn't manage to do that. Will do it later this afternoon.

students/mattmaeda/session08/sparse_array.py
```diff
@@ -0,0 +1,74 @@
...
+        if index > self.__length - 1:
+            raise IndexError
+        else:
+            for i in range(index, (self.__length - 1)):
+                if (i+1) in self.__array:
+                    self.__array[i] = self.__array[i+1]
+
+            self.__array.pop((self.__length - 1), None)
+            self.__length -= 1
+
```
codedragon, line 51: You have a bug:
```
In [2]: sparse = SparseArray([1, 0, 0, 1, 3, 9, 0])

In [3]: sparse[0]
Out[3]: 1

In [4]: sparse[1]
Out[4]: 0

In [5]: del sparse[0]

In [6]: sparse
Out[6]: [1,0,1,3,9,9]
```

students/mattmaeda/session08/sparse_array.py
```diff
@@ -0,0 +1,74 @@
...
+        self.__index += 1
+
+        if self.__length >= self.__index:
+            if (self.__index - 1) in self.__array:
+                return self.__array[(self.__index - 1)]
+            return 0
+
+        else:
+            raise StopIteration
+
```
codedragon, line 68: This is subtle, but your iterator does not work quite like other python iterators:
```
In [6]: sparse
Out[6]: [1,0,1,3,9,9]

In [7]: test = iter(sparse)

In [8]: next(test)
Out[8]: 1

In [9]: next(test)
Out[9]: 0

In [10]: next(test)
Out[10]: 1

In [11]: for i in test:
    ...:     print(i)
    ...:
1
0
1
3
9
9

In [12]: new = [1, 2, 3, 4, 5]

In [13]: test2 = iter(new)

In [14]: next(test2)
Out[14]: 1

In [15]: next(test2)
Out[15]: 2

In [16]: for i in test2:
    ...:     print(i)
    ...:
3
4
5
```

PR #136, opened 2017-12-01
First pull request in a while

Topcat7:
Don't feel obligated to comment on everything, I would rather you focus on html render and the circle lab if possible.

Thank you


PythonCHB:
looking really good!

PR #137, opened 2017-12-02
Completed html renderer

Eowyn42:
Completed all steps and am rendering lovely HTML. run_html_renderer.py produces great results, tests have I believe adequate coverage of relevant items, and in general I'm very happy with this. Oh and I refactored a few things to reduce repetition. If any review happens, it would be on the test coverage of steps 4-8 tasks. Thank you.

### DIFF
--- a/students/eowyn/session07/html-rendering/html_render.py
+++ b/students/eowyn/session07/html-rendering/html_render.py
@@ -4,11 +4,12 @@ class Element():
     tag = 'html'
     extra_indent = 4 * ' '
 
-    def __init__(self, content=None):
+    def __init__(self, content=None, **kwargs):
         if content is None:
             self.content = []
         else:
             self.content = [content]
+        self.attributes = kwargs  # a dictionary w/all the keyword args
 
     def append(self, content):
         # Collect additional objects alongside existing objects
@@ -18,15 +19,40 @@ class Element():
         # Pretty-print content and tags to file.
         # For objects, pretty-print all the content and tags
         # they contain to the file. Nest tags with indentation.
-
-        out_file.write('\n' + current_ind + '<' + self.tag + '>')
+        self.render_open_tag(out_file, current_ind)
         for each in self.content:
             try:
                 each.render(out_file, current_ind + self.extra_indent)
             except AttributeError:
                 out_file.write('\n' + current_ind +
                                self.extra_indent + str(each))
-        out_file.write('\n' + current_ind + '</' + self.tag + '>')
+        out_file.write('\n')
+        close_tag = '{}</{}>'.format(current_ind, self.tag)
+        out_file.write(close_tag)
+
+    def render_open_tag(self, out_file, current_ind):
+        # This code was repeated so I refactored it!
+        # Get and write an open tag to out_file
+        open_tag = self.get_open_tag()
+        out_file.write('\n')
+        out_file.write(current_ind)
+        out_file.write(open_tag)
+
+    def get_open_tag(self):
+        # This code was repeated so I refactored it!
+        # Construct the open tag with its attributes, if present
+        open_tag = '<{}'.format(self.tag)
+        open_tag += self.get_tag_attributes()
+        open_tag += ">"
+        return open_tag
+
+    def get_tag_attributes(self):
+        # This code was repeated so I refactored it!
+        # Construct string of any/all attributes
+        att_str = ""
+        for att, val in self.attributes.items():
+            att_str += ' {}="{}"'.format(att, val)
+        return att_str
 
 
 class Body(Element):
@@ -45,7 +71,6 @@ class Html(Element):
         # For objects, pretty-print all the content and tags
         # they contain to the file. Nest tags with indentation.
         # For html files, include a top level DOCTYPE string.
-
         out_file.write('<!DOCTYPE html>')
         Element.render(self, out_file, ind)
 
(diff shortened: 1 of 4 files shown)


Eowyn42:
Finished the circle class, too. I included subtraction, true division, floor division, modulo and reflective operators for add/subtract. I did not like the idea of circles with negative radii so I implemented the abs() unary operator, which I then used in the subtraction and r-sub operators. 

Question 1: do I really need to define a new radius and return a new circle with that radius? I wanted to just work with self, other, but odd things were happening
Question 2: there's repeated code with the reflective operators, is that avoidable?

I hope to attend office hours & during that time implement __pow__ and other ones that make sense

students/eowyn/session08/circle.py
```diff
@@ -65,6 +65,10 @@ def __sub__(self, other):
         newrad = self.radius - other.radius
         return abs(Circle(newrad))
 
+    def __mod__(self, other):
```
PythonCHB: not sure how useful mod is for circles .. but what the heck for completeness' sake :;)

but couldn't you use the built-in mod operator: % ?

students/eowyn/session07/html-rendering/html_render.py
```diff
@@ -18,15 +19,40 @@ def render(self, out_file, current_ind=""):
...
                 each.render(out_file, current_ind + self.extra_indent)
             except AttributeError:
                 out_file.write('\n' + current_ind +
                                self.extra_indent + str(each))
-        out_file.write('\n' + current_ind + '</' + self.tag + '>')
+        out_file.write('\n')
+        close_tag = '{}</{}>'.format(current_ind, self.tag)
+        out_file.write(close_tag)
+
+    def render_open_tag(self, out_file, current_ind):
```
PythonCHB, line 33: great idea to pull this out to a separate method.

but I think it's a bit cleaner to have this method return the entire tag as a string, and then actually write it to the out_file in the render method. That keeps all the actual file writing in one place.

students/eowyn/session07/html-rendering/html_render.py
```diff
@@ -18,15 +19,40 @@ def render(self, out_file, current_ind=""):
...
+
+    def get_open_tag(self):
+        # This code was repeated so I refactored it!
+        # Construct the open tag with its attributes, if present
+        open_tag = '<{}'.format(self.tag)
+        open_tag += self.get_tag_attributes()
+        open_tag += ">"
+        return open_tag
+
+    def get_tag_attributes(self):
```
PythonCHB, line 49: very nice!

students/eowyn/session07/html-rendering/html_render.py
```diff
@@ -59,12 +84,73 @@ class OneLineTag(Element):
...
     tag = 'title'
+
+
+class SelfClosingTag(Element):
+    # render tags like <hr /> and <br />
+    # render just one tag and any attributes
+
+    def render(self, out_file, current_ind=""):
+        # Render just a tag and any attributes, ignore contents
+        self.render_open_tag(out_file, current_ind)
```
PythonCHB, line 104: since all this doing is calling render_open_tag(), and that is overidden this this subclass, you may as well put the code in there right in here.

unless you think there are other subclasses that might use this get_open_tag method.

students/eowyn/session07/html-rendering/html_render.py
```diff
@@ -59,12 +84,73 @@ class OneLineTag(Element):
...
+    tag = 'br'
+
+
+class A(OneLineTag):
+    # Make anchor tags for hyperlinks, like this:
+    # <a href="http://google.com"> link to google </a>
+    # Instructions said subclass from Element but we want it on one line
+    tag = 'a'
+
+    def __init__(self, link=None, content=None, **kwargs):
```
PythonCHB, line 128: perfect! remarkably easy, isn't it?

students/eowyn/session07/html-rendering/html_render.py
```diff
@@ -59,12 +84,73 @@ class OneLineTag(Element):
...
 class Title(OneLineTag):
     tag = 'title'
+
+
+class SelfClosingTag(Element):
+    # render tags like <hr /> and <br />
+    # render just one tag and any attributes
+
+    def render(self, out_file, current_ind=""):
+        # Render just a tag and any attributes, ignore contents
```
PythonCHB, line 103: nice! but maybe you should raise an error if there are any contents -- probably in the __init__ -- that would be better than simply ignoring it.

students/eowyn/session08/circle.py
```diff
@@ -0,0 +1,131 @@
...
+        # reflective multiply... repeats code??
+        newrad = self.radius * factor
+        return Circle(newrad)
+
+    def __truediv__(self, factor):
+        newrad = self.radius / factor
+        return Circle(newrad)
+
+    def __rtruediv__(self, factor):
+        # reflective division... repeats code??
```
PythonCHB, line 49: it's reveresed -- is it giving the right answer? I woulndn't expect so!

but if it is repeated, you can re-use the previous definition:

def __rtrudiv__ = __truediv__

pretty cool, eh?

students/eowyn/session08/circle.py
```diff
@@ -0,0 +1,131 @@
...
+        return Circle(newrad)
+
+    def __idiv__(self, factor):
+        newrad = self.radius / factor
+        return Circle(newrad)
+
+
+#  ---------------- UNARY ARITHMETIC OPERATORS -----------------------
+
+    def __abs__(self):
```
PythonCHB, line 126: another one that may not make sense....

PythonCHB (approved):
This all looks great --nice work!

PR #138, opened 2017-12-03
Finished HTML Renderer and Circle Class

chops77:


students/Chao/session08/html_render.py
```diff
@@ -0,0 +1,115 @@
...
+            file_obj.write(' ' + at + '="' + val + '"')
+        file_obj.write('>')
+        for each in self.content:
+            newind = ind + self.indent
+            if type(each) == str:
+                file_obj.write(each)
+            else:
+                each.render(file_obj, newind)
+        file_obj.write('</' + self.tag + '>\n')
+
```
codedragon, line 67: This is so close to the code you have for Element, I feel like there should be a way to come up with a render method that could work for either.

students/Chao/session09/mailroom_oo.py
```diff
@@ -0,0 +1,16 @@
...
+        self.amount = [amount]
+
+    def new_donation(self, amount):
+        self.amount.append(amount)
+
+    @property
+    def total_donation(self):
+        return sum(self.amount)
+
+
```
codedragon, line 15: Good start! Now you need a class to store and manipulate all of the donor objects.

codedragon (approved):
Great work so far! :) Hardly had any comments to make.

PR #139, opened 2017-12-04
step 8: code and test for meta are not working

HiroTestStudent:
Hi Chris and Maria,

Thank you for help during office hours this weekend. I wrote the codes thru step 8 but I got stuck with last script for <meta charset="UTF-8" />. Instruction says we should inherit from SelfClosingTag class but I got confused in the middle of coding. Is it supposed to be inherit from OneLineTag class? 

Thanks,
Hiro

### DIFF
--- a/students/hirot_directory/session10/oo_mailroom_hiro/mailroom.py
+++ b/students/hirot_directory/session10/oo_mailroom_hiro/mailroom.py
@@ -1,0 +1,181 @@
+#!/usr/bin/env python3
+
+'''
+Hiro
+'''
+
+import sys
+import math # from chris
+from textwrap import dedent # from chris
+
+
+def add_donor():
+    donor_name = str(input("Full Name: "))
+    return donor_name
+
+
+def add_donation():
+    new_donation = int(input("How much do you want to donate?: "))
+    return new_donation
+
+
+def select_submenu():
+    choice = int(input("Select from one of these options:\n"
+                              "(1) Show a list\n"
+                              "(2) Enter a donation\n"  
+                              "(3) Exit\n"
+                              ))
+    return choice
+
+
+def select_mainmenu():
+    answer = int(input("Select from one of these options:\n"
+                                   "(1) Send a Thank you\n"
+                                   "(2) Create a Report\n"
+                                   "(3) Send letters to everyone\n"
+                                   "(4) quit\n"
+                                   ))
+    return answer
+
+
+def select_quit():
+    selection = (input("Are you sure to quit? (Y or N) To return to main menu, please select 'N'.: ")).upper()
+    return selection
+
+
+def mainloop(self):
+    while True:
+        try:
+            answer = self.select_mainmenu()
+            self.dict_answer[answer]()
+
+        except ValueError as e:
+            print("Error happened. Error is ", type(e), "Please type 1, 2 or 3")
+        except KeyError as k:
+            print("Error happened. Error is ", type(k), "Please type 1, 2 or 3")
+
+
+class DonorDB(): #donors and donations
+
+
+
+
+
+
+class Donor(): # a donor and a donation
+
+    def __init__(self):
+        self.donor_names = ["William Gates III", "Mark Zuckerberg", "Jeff Bezos", "Paul Allen"]
+        self.donations = [[200, 500], [2000, 3000], [4000, 5000], [100, 150]]
+        self.donor_dict = {self.donor_names:self.donations for (self.donor_names, self.donations) in zip(self.donor_names, self.donations)}
+        self.dict_answer = {1: self.thank_you_loop, 2: self.create_report, 3: self.send_letters, 4: self.quit}
+        self.dict_choice = {1: self.print_list, 2: self.make_donation, 3: self.quit}
+
+
+    def print_list(self):
+
+        print("{msg1: <50}| {msg2: <12}| " \
+
+              "{msg3:<10}| {msg4: <12}".format(msg1="Donor Name",
+
+                                               msg2="Total Given",
+
+                                               msg3="Num Gifts",
+
+                                               msg4="Average Gift"))
+        for d in self.donor_dict:
+            t = sum(self.donor_dict[d])
+            n = len(self.donor_dict[d])
+            a = t / n
+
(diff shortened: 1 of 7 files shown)


PythonCHB:
no, it's correct. the meta tag:

`<meta charset="UTF-8" />`

is a self-closing tag with a single attribute.

so your SelfClosingTag class should support attributes, just like the others.

HiroTestStudent:
Hi Chris, 

OK, then I am suspecting that I might need to refactor SelfClosingTag class. Please let me discuss with you after the class (or before the class if any time). 

I spent too much time on html_render assignment so I couldn't work on circle assignment...

Thanks,
Hiro

HiroTestStudent:
Hi Chris and Maria,

I just uploaded the semi-final version. It has codes through step 8. As we just spoke, step 7 needs to be more "DRY", but code is working.

Thanks,
Hiro

HiroTestStudent:
Hi Maria and Chris,

I just pushed OOP mailroom.py. The code is working (with Maria's help during office hours today). I forgot to update the test_mailroom.py so I will work on test script and push it later...

Thanks,
Hiro

PythonCHB:
tsk tsk .. you shold have done the tests FIRST :-)

-CHB

students/hirot_directory/session10/oo_mailroom_hiro/mailroom.py
```diff
@@ -0,0 +1,212 @@
...
+        else:
+            return mainloop()
+
+    except ValueError as e:
+        print("Error happened. Error is ", type(e), "Please type 1, 2 or 3")
+    except KeyError as k:
+        print("Error happened. Error is ", type(k), "Please type 1, 2 or 3")
+
+
+def make_sample_db():
```
PythonCHB, line 77: now that you have a DonorDB class, this function should return one of those!

it could start by making an empty DonorDB, then loop through the names and donations, call add_donor on the DonorDB, then add the donations to the donor.

students/hirot_directory/session10/oo_mailroom_hiro/mailroom.py
```diff
@@ -0,0 +1,212 @@
...
+
+    # return [Donor("William Gates III", [653772.32, 12.17]),
+    #         Donor("Jeff Bezos", [877.33]),
+    #         Donor("Paul Allen", [663.23, 43.87, 1.32]),
+    #         Donor("Mark Zuckerberg", [1663.23, 4300.87, 10432.0]),
+    #         ]
+
+
+def find_donor(name):
+    if name in make_sample_db().sample_db.keys():
```
PythonCHB, line 97: if name in make_sample_db().sample_db() would do the dame thing.

But this should be finding teh donor IN this DonorDB!!!

if name in self.all_donors:

should do it.

and you don't want to return the name, you want to return the donor!

to make it more compact:

`return self.all_donors.get(name, None)`

That will give you the donor associated with the name if it's there, and None otherwise.

students/hirot_directory/session10/oo_mailroom_hiro/mailroom.py
```diff
@@ -0,0 +1,212 @@
...
+
+              "{msg3:<10}| {msg4: <12}".format(msg1="Donor Name",
+
+                                               msg2="Total Given",
+
+                                               msg3="Num Gifts",
+
+                                               msg4="Average Gift"))
+        for donor in self.all_donors:
+            for key in donor.donor_dict:
```
PythonCHB, line 137: the Donor object doesn't (or shouldn't) have a donor_dict -- it has a name and donations....

students/hirot_directory/session10/oo_mailroom_hiro/mailroom.py
```diff
@@ -0,0 +1,212 @@
...
+        print("{msg1: <50}| {msg2: <12}| "
+
+              "{msg3:<10}| {msg4: <12}".format(msg1="Donor Name",
+
+                                               msg2="Total Given",
+
+                                               msg3="Num Gifts",
+
+                                               msg4="Average Gift"))
+        for donor in self.all_donors:
```
PythonCHB, line 136: This will give you all the keys (names) you want the donor objects.

You want: for donor in `self.all_donors.values():`

students/hirot_directory/session10/oo_mailroom_hiro/mailroom.py
```diff
@@ -0,0 +1,212 @@
...
+
+
+class Donor():
+
+    def __init__(self, donor_name, donations=()):
+        self.name = donor_name.strip()
+        self.donations = list(donations)
+
+    def make_donation(self, new_donation):
+
```
PythonCHB, line 189: you may want to add some code here to check that you have a reasonable donation value -- call flost() onitto makeit a number, and maybe check that it's > 0.0 ...

students/hirot_directory/session10/oo_mailroom_hiro/test_mailroom.py
```diff
@@ -0,0 +1,182 @@
...
+    donor = db.add_donor(name)
+    donor.make_donation(300)
+    assert donor.name == "Fred Flintstone"
+    assert donor.last_donation == 300
+    # assert db.find_donor(name) == donor
+
+
+def test_find_donor():
+    """ checks a donor that is there, but with odd case and spaces"""
+    donor = mailroom.find_donor("jefF beZos ")
```
PythonCHB, line 72: does this work? I did a nifty normalization so all capitalization doesn't matter when searching for a donor. But you may want to start with having to ask for the exact name.

and when this test passes, make sure it's using the DonorDB find_donor() method!

PythonCHB (commented):
Making progress, but make sure to step back and think about what you are doing. The Object oriented refactor is supposed to encapsulate code and data together where they belong.

Each Donor object only knows about itself. it doesn't know anything about any other donors.

The DonorDB knows about the entire collection of donors, but it only knows the interface for donor objects that you provide -- it doesn't know how the Donor is storing the donations, for example, but it shoud hav emthods (or properties) for:

donor.total_dontations
donor.last_donation
donor.average_donation

for instance.

You will be able to re-factor your old code for the actual functionality, but it will be in different places, and should not refer to old data structures.

PR #140, opened 2017-12-05
Session 08

kathrynegan:
Circle.py, SparseArray.py, and their test suites.

There's a question at the top of Circle that I'd like to get your thoughts on. Any advice you can provide about making the classes more functional, stable, and reusable would be great. In addition, I'd appreciate any input about how to make my testing more robust and generalizable.

SparseArray was a great challenge - and fun! Thanks for providing the optional assignment, I know it means more work for you.


kathrynegan:
Wanted to update code, closing this pull request. Will pull again after commits.

PR #141, opened 2017-12-05
mailroom, classes and render files

Ginsusamurai:


PythonCHB:
I only see mailroom files.

Did you mean to commit html_render code as well?

And do you need mailroom reviewed?

Ginsusamurai:
Guess I had some commits missing from my last PR.

Finished up my class-based Mailroom attempt (mailroom4.py) and added that to this updated PR.

students/morgan/Mailroom/mailroom4.py
```diff
@@ -0,0 +1,215 @@
...
+        return sum(self.donations)
+
+    @property    
+    def maxdonation(self):
+        return max(self.donations)
+
+        
+
+    @property
+    def namelength(self):
```
PythonCHB, line 85: not  muych point to this -- a_donor.name is expected to be a string -- so if you want to know how long it is, you can simply so:

len(a_donor.name)

you want a property to "hide" implementation details -- but the name attribute being a string is part of the API, not an implementation detail.

PythonCHB: The other properties are great -- they hide (or encapsulate) the fact that the donations are stored as a simple list.

students/morgan/Mailroom/mailroom4.py
```diff
@@ -0,0 +1,215 @@
...
+class Storage:
+
+
+    def __init__(self):
+        self.donors = []
+
+
+    #need methods to add donors, add donation
+    # @classmethod
+    def add_donor(self, thing):
```
PythonCHB, line 110: I'd use a more meaningful name than "thing" -- this really should only  be called with a Donor object, yes?

students/morgan/Mailroom/mailroom4.py
```diff
@@ -0,0 +1,215 @@
...
+        self.donors.append(thing)
+
+
+
+    def add_donation(self, name, new_donation):
```
PythonCHB, line 115: perhaps you should only use the add_donation() method of the donor object, rather than finding a donor and adding in one step.

in any case, you want a "find_a_donor" method in this class, which you could use if you want find and add at the same time.

students/morgan/Mailroom/mailroom4.py
```diff
@@ -0,0 +1,215 @@
...
+        ind_temp = False
+
+        for i, x in enumerate(self.donors):
+            if (x.name).lower() == (name).lower():
+                ind_temp = i
+                self.donors[i].donations.append(new_donation)
+                print(i)
+                print("\nA donation of {} has been added to the history of {}".format(float(new_donation),name))
```
PythonCHB, line 123: you don't want any print calls inside these classes -- that's up to the user-interaction code.

PythonCHB (approved):
Great first pass at an OO approach.

PR #142, opened 2017-12-05
adding circle.py and test..up through step 5

ea8606:
put a step in the test file to test for an attribute error... is this the pythonic way to test for errors?

### DIFF
--- a/students/EricAdams/session09/circle.py
+++ b/students/EricAdams/session09/circle.py
@@ -1,0 +1,38 @@
+#! /usr/bin/env python
+# circle.py
+# A Circle can be defined by either specifying the radius or the diameter,
+# and the user can query the circle for either its radius or diameter.
+
+# Other abilities of a Circle instance:
+
+# Compute the circle’s area
+# Print the circle and get something nice
+# Be able to add two circles together
+# Be able to compare two circles to see which is bigger
+# Be able to compare to see if there are equal
+# (follows from above) be able to put them in a list and sort them
+
+import math
+
+
+class Circle():
+    def __init__(self, radius):
+        self.radius = radius
+    _diameter = 0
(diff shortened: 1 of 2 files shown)


students/EricAdams/session09/test_circle.py
```diff
@@ -0,0 +1,49 @@
...
+    circle_obj = circle.Circle(r)
+    assert circle_obj.area == math.pi * r**2
+
+
+def test_circle_area_setter():
+    r = 4
+    circle_obj = circle.Circle(r)
+    print(circle_obj.area)
+    with pytest.raises(Exception) as excinfo:
+        circle_obj.area = 10
```
PythonCHB, line 45: This can be much simpler :


with pytest.raises(AttributeError):
    circle_obj.area = 10

Thst’s It ;-)

students/EricAdams/session09/circle.py
```diff
@@ -0,0 +1,38 @@
...
+# Be able to compare to see if there are equal
+# (follows from above) be able to put them in a list and sort them
+
+import math
+
+
+class Circle():
+    def __init__(self, radius):
+        self.radius = radius
+    _diameter = 0
```
PythonCHB, line 21: You really don’t want to be storing the diameter. You are re-calculating it whenever you get it in any case.

And if you Store it, it could get out of sync with radius...

students/EricAdams/session09/circle.py
```diff
@@ -0,0 +1,38 @@
...
+    _area = 0
```
PythonCHB, line 22: Or area....

PythonCHB (approved):
Looking good. Try simply removing _area and _diameter, and your tests should still pass.

PR #143, opened 2017-12-06
'submitting html_render.py and test_html_render.py for review

zandra2:
I commented out most of the assert test for 'test_render_body' and 'test_render_para' because I didn't want you to get a failed msg when you run it.  However, I am not understanding why I get an AttributeError: 'Element' object has no attribute 'startswith' or 'endswith' when these two classes inherited from Element's and the 'test_render' passed the test.

Also, for the 'test_render_html',  I get a TypeError: argument of type 'Html' is not iterable. Why is that?

Thank you.

### DIFF
--- a/students/zandra_eng/session07/test_html_render.py
+++ b/students/zandra_eng/session07/test_html_render.py
@@ -1,0 +1,88 @@
+#!/usr/bin/env python3
+
+
+from html_render import (Element, Html, P, H, Body, Title)
+
+def test_new_element():
+    """elements can be initialized"""
+    el_obj = Element()
+    el_obj = Element('content')
+
+def test_add_content():
+    el_obj = Element('content')
+    assert el_obj.content == [('content')]
+
+def test_add_empty_str():
+    el_obj = Element('')
+    assert el_obj.content == ['']
+
+def test_append_str():
+    el_obj = Element('str, str, str')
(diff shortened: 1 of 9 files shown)


PythonCHB:
"AttributeError: 'Element' object has no attribute 'startswith' or 'endswith' when these two classes inherited from Element's and the 'test_render' passed the test.'

startwith and endswith are string methods -- in your tests, you should be using these to test the results of rendering, not calling them on the elements themselves.

students/zandra_eng/session07/test_html_render.py
```diff
@@ -0,0 +1,88 @@
...
+    # assert contents.endswith('</body>')
+    # assert something in contents
+    # assert something_more in contents
+
+def test_render_para():
+    something = 'str, str, str'
+    el_obj = Element(something)
+    something_more = 'more, more, more'
+    el_obj.append(something_more)
+    contents = Element('str str str more more more')
```
PythonCHB, line 71: you need to actually render the Element in order to test the results -- in my example code, I have a functiont hat will render the element, and return the results as a string.

PythonCHB:
Hmm -- you've got a bunch of other folks' files in her now -- as you know.

So we'll see if we can get it cleaned up in office hours...

PR #144, opened 2017-12-06
ClassMailroom

combatraffi:
Lastest push contains class based mailroom.  Still needs work but its basic flow and functionality is there.  Testing to follow.


students/Dkuchan/session6/classessheet.py
```diff
@@ -0,0 +1,51 @@
...
+        lastfirstname = '{}, {}'.format(self.last_name, self.first_name)
+        return lastfirstname
+
+    def appenddonations(self, newdon):
+        ''' Adds a donation to the donations list
+            Updates the average donations
+            Updates the max donation '''
+        self.donations.append(newdon)
+        self.updatecalcs()
+
```
codedragon, line 37: I think your Donor class would be easer if each instance represented a single donor, and then these donors were put into a database or another class. There is quite a bit of awkwardness here that I think could be sorted out by separating individual donor info from the group. The code seems to be broken, btw, because you increase numdonors when you initialize, which means Dan is in position 1 of your donor list (with None still in position 0), but when you add Abbey, she is 2 and the donor list you created only has 2 positions, so she cannot be added to your list.

students/Dkuchan/session08/frcscouter.py
```diff
@@ -0,0 +1,145 @@
+# FRC SCOUTER
+# frcscouter.py
+
+class Competition():
+    # this class holds information specific to a certain competition
+    def __init__(self, cname):
+        self.compname = cname
+        self.teamroster = []
+
```
codedragon, line 9: If you just have an init with a couple of variables, you might as well use a dictionary instead.

students/Dkuchan/session08/frcscouter.py
```diff
@@ -0,0 +1,145 @@
...
+    def __init__(self, mnumber)
+        self.teamnumber
+        self.matchnum = mnumber
+        self.alliance
+        self.winloss
+        self.pressure
+        self.gears
+        self.autogear
+        self.replaylink
+
```
codedragon, line 22: Actually, all of these classes should just be dictionaries.

students/Dkuchan/session07/classmailroom.py
```diff
@@ -0,0 +1,263 @@
...
+        return(max(self.donations))
+
+    @property
+    def averagedon(self):
+        return(self.totaldonations / len(self.donations))
+
+    @property
+    def namelen(self):
+        return(len(str(self.fullname)))
+
```
codedragon, line 35: This shouldn't be necessary. self.fullname should always be a string, and so len(self.fullname) should be sufficient. Also, this really doesn't need to be a property, since it is only used in one place, and the user won't care about it. Just find it directly in the longestname function.

students/Dkuchan/session07/classmailroom.py
```diff
@@ -0,0 +1,263 @@
...
+donorlist.append(Donor('Chris Jaeger'))
+
+def longestname():
+    # Finds the longest name and returns its length.
+    length = 0
+    for i in range(0, len(donorlist)):
+        if donorlist[i].namelen >= length:
+            length = donorlist[i].namelen
+    return length
+
```
codedragon, line 63:
```
length = 0
for donor in donorlist:
    if len(donor.fullname) > length:
        length = len(donor.fullname)
```
or more compact:
```
return max(len(donor.fullname) for donor in donorlist)
```

students/Dkuchan/session07/circles.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+class Sphere(Circle):
+
+    def volume(self):
+        return (4 / 3 * pi * self.radius ** 3)
+
+    def area(self):
+        return (4 * pi * self.radius ** 2)
+
```
codedragon, line 38: Since area is a property in the Circle object, and volume is very similar, they should both be properties here too.

codedragon (approved):
This is a lot for one code review. Please next quarter submit more frequently, or add comments about what specifically you would like us to look at. I just made comments on stuff for 7 and later. I would love to see tests!

PR #146, opened 2017-12-07
Session 08

kathrynegan:


students/kegan/session10/MailroomTools.py
```diff
@@ -0,0 +1,489 @@
...
+            donations (args) : donations as arguments
+        """
+        self._name = self.clean_name(name)
+        self._donations = self.intake_donations(*donations)
+        if not self._donations:
+            raise ValueError(
+                '{} must have at least one donation > 0'.format(self._name))
+
+    @property
+    def name(self):
```
PythonCHB, line 26: you generally don't want a property that simply returns a "private" attribute -- you can just keep name as a regular attribute.

students/kegan/session10/MailroomTools.py
```diff
@@ -0,0 +1,489 @@
...
+import functools
+
+
+@functools.total_ordering
+class Donor:
+
+    def __init__(self, name, *donations):
+        """ Initializes Donor object with given name and
+        list of donations. Will not accept donations <= 0.
+        Will raise ValueError if there are no donations > 0
```
PythonCHB, line 13: Why not be able to create an empty one?

students/kegan/session10/MailroomTools.py
```diff
@@ -0,0 +1,489 @@
+"""
+Kathryn Egan
+"""
+import functools
+
+
+@functools.total_ordering
+class Donor:
+
+    def __init__(self, name, *donations):
```
PythonCHB, line 10: hmm -- this means that to pass in a list of donations, they need to unpack it:

Donor("Fred, *list_of_donations).

It does mean they can pass in a single donation, which is nice.

I"d probably take a sequence of donations as a single parameter instead, but his is kind of nifty.

students/kegan/session10/MailroomTools.py
```diff
@@ -0,0 +1,489 @@
...
+    @donations.setter
+    def donations(self, donations):
+        """ Sets donations.
+        Args:
+            donations (args) : donations as arguments
+        """
+        self._donations = self.intake_donations(*donations)
+
+    @name.setter
+    def name(self, name):
```
PythonCHB, line 74: ahh -- you have the getter so you can have a setter - that does make sense. 

But you want to put you properties setters right next to the getter -- that way it's obvious when you read the code what's going on.

students/kegan/session10/MailroomTools.py
```diff
@@ -0,0 +1,489 @@
...
+        """ Cleans given name.
+        Args:
+            name (str) : donor name
+        Returns:
+            str : cleaned name
+        """
+        return ' '.join(name.split()).title()
+
+    @staticmethod
+    def intake_donations(*donations):
```
PythonCHB, line 92: you can use an underscore to indicate the a method is designed for internal use as well, This might be a good place to do that.

students/kegan/session10/MailroomTools.py
```diff
@@ -0,0 +1,489 @@
...
+
+    def add(self, *donations):
+        """ Adds passed donations to this donor.
+        Args:
+            donations (args) : donations as arguments
+        """
+        donations = self.intake_donations(*donations)
+        self._donations.extend(donations)
+
+    def __str__(self):
```
PythonCHB, line 149: you want to be a bit careful here -- if there are a LOT of donations, this could get really big.

I like to have a string that provides a summary:

The Name, with N donations totally N dollars, or something like that. or have it show you the last N donations, chopping off the extras if there are any.

students/kegan/session10/MailroomTools.py
```diff
@@ -0,0 +1,489 @@
...
+    def __repr__(self):
+        """ Returns representation of this donor.
+        Returns:
+            str : representation of this donor
+        """
+        donations = ', '.join([
+            str(d) for d in self.donations])
+        return 'Donor("{}", {})'.format(self.name, donations)
+
+    def __eq__(self, other):
```
PythonCHB, line 167: careful here -- you probably NEVER want two donors to be equal! what are the odds that two donors would have the same name and donation history.

So I'd be disinclined to define this.

If you want the sortable, you can define __lt__ alone, and that will work.

students/kegan/session10/MailroomTools.py
```diff
@@ -0,0 +1,489 @@
...
+    def __lt__(self, other):
+        """ Returns whether this donor is less than
+        other. Compares name and donations.
+        Args:
+            other (Donor) : donor to compare
+        Returns:
+            bool : True if this donor is less than other
+        """
+        return (
+            self.name < other.name and
```
PythonCHB, line 188: careful here.

by using "and", one donor has to have a name less than AND fewer donations. the name less than is good -- that's alphabetical order.

But also requiring the donations to be less means you can have odd results if one name is greater than the other, but the donations are less. I prototyped this, add this is what I get:

```
In [33]:     d1 = Donor("Abe", [2])
    ...:     d2 = Donor("Bob", [1])
    ...: 

In [34]: d2 < d1
Out[34]: False

In [35]: d1 < d2
Out[35]: False
```

What you probably want is to check the name, and only if the names are equal, then test the donations. you could do this with a bit of logic, but the easiest way is probably to put them in a tuple:

return (self.name, self.donations) < (other.name, other.donations)

tuples are sorted in order of the items -- only if the first item is the same does the second one get checked.

Also, probably sorting by total donations would make more sense, as if you sort on teh list, the firsdt one dominates:

```
In [36]: [1,2,3] < [10,20,30]
Out[36]: True
```
all good.

```
In [37]: [1,2000] < [10,20,30]
Out[37]: True
```
maybe not what you'd want

Finally, of the only reason to implement comparisons is to get a good sort order, implementing a sort_key function that can be used will result in faster sorting:

```
    @staticmethod
    def sort_key(self):
        return (self.name, self.donations)
```

Then: 

sorted(list_of_donors, key=Donor.sort_key)

students/kegan/session10/MailroomTools.py
```diff
@@ -0,0 +1,489 @@
...
+            self.donations < other.donations)
+
+    def __contains__(self, donation):
```
PythonCHB, line 191: is this useful?

not bad practice, but there are a lot of magic methods, you generally don't implement ones that aren't useful.

Also, you have exposed a donations list, s if someone rally wanted to know if a particular donation amount was in there:

`amount in donor.donations`

would work, and not require them knowing anything "private" about the implementation.

Though you are right that if getting the donations list was expensive (Pulling for a DB, etc) than it might be good to have a __contains__ method that could do an efficient DB query.

students/kegan/session10/MailroomTools.py
```diff
@@ -0,0 +1,489 @@
...
+        Args:
+            donation (int or float) : donation to search for
+        Returns:
+            bool :
+                True if donation is in donation list
+                False otherwise
+        """
+        return donation in self.donations
+
+    def multiply(self, factor, min_donation=None, max_donation=None):
```
PythonCHB, line 203: very nicely done!

students/kegan/session10/MailroomTools.py
```diff
@@ -0,0 +1,489 @@
...
+        return Donor(self.name, *donations)
+
+    @staticmethod
+    def apply_factor(d, factor, min_donation, max_donation):
+        within_min = True if min_donation is None else d >= min_donation
+        within_max = True if max_donation is None else d <= max_donation
+        return d * factor if within_min and within_max else d
+
+
+@functools.total_ordering
```
PythonCHB, line 215: this is for implementing total ordering on individual items -- are you ever going to need to compare different DonorLists ?

students/kegan/session10/MailroomTools.py
```diff
@@ -0,0 +1,489 @@
...
+class DonorList:
+
+    def __init__(self, *donors):
+        """ Initializes list of donors.
+        Args:
+            donors (args) : donors as arguments
+        """
+        self._donors = [donor for donor in donors]
```
PythonCHB, line 223: `list(donors)` is a bit more efficient and obvious.

students/kegan/session10/MailroomTools.py
```diff
@@ -0,0 +1,489 @@
...
+            except IndexError:
+                continue
+            else:
+                str_donations = [
+                    item.strip().strip('$') for item in line[1:]]
+            donations = []
+            for d in str_donations:
+                try:
+                    d = float(d)
+                except TypeError:
```
PythonCHB, line 259: if d is a string with an ivalid value, you will get a VAlueError, not a TypeError.

you'll get a TypeError if it is a type that float( never knows how to convert.

so this should probably be:

`except TypeError, ValueError:`

Also -- do you want to simply ignore invalid input? probably not.

students/kegan/session10/MailroomTools.py
```diff
@@ -0,0 +1,489 @@
...
+    @property
+    def donors(self):
+        """ Returns donors.
+        Returns:
+            list : list of donors
+        """
+        return self._donors
+
+    @donors.setter
+    def donors(self, donors):
```
PythonCHB, line 282: do you ever have a reason to set the entire donor list?

PythonCHB: if so, some validation would probably be good.
and if you have a property that simply gets and sets a value -- you don't need a property.

students/kegan/session10/MailroomTools.py
```diff
@@ -0,0 +1,489 @@
...
+        self._donors = donors
+
+    def add(self, donor):
+        """ Adds given donor to donor list.
+        Args:
+            donor (Donor) : donor to add to donor list
+        """
+        self._donors.append(donor)
+
+    def sort_by(self, key, low_to_high=False):
```
PythonCHB, line 296: This is really nice!

students/kegan/session10/MailroomTools.py
```diff
@@ -0,0 +1,489 @@
...
+        functions = {
+            'total': lambda donor: donor.total,
+            'average': lambda donor: donor.average,
+            'num': lambda donor: donor.num}
+        if key not in functions:
+            raise KeyError('Invalid argument {}'.format(key))
+        return DonorList(*sorted(
+            self.donors, key=functions[key], reverse=not low_to_high))
+
+    def __getitem__(self, name):
```
PythonCHB, line 314: slick!

students/kegan/session10/MailroomTools.py
```diff
@@ -0,0 +1,489 @@
...
+        """ Returns whether this donor list has all the
+        same donors as the other donor list.
+        Args:
+            other (DonorList) : DonorList to compare
+        Returns:
+            bool : True if both lists share donors, False otherwise
+        """
+        return list(self) == list(other)
+
+    def __lt__(self, other):
```
PythonCHB, line 385: why?

or for practice's sake...

students/kegan/session10/MailroomTools.py
```diff
@@ -0,0 +1,489 @@
...
+        report.append('-' * (report_width - 1))
+        for donor in self.sort_by('total'):
+            row = [
+                form(yld(donor), width)
+                for (_, width, form, yld) in columns]
+            report.append(' '.join(row))
+        return '\n'.join(report)
+
+    @staticmethod
+    def name(name, width):
```
PythonCHB, line 449: maybe a regular method that uses self.name?

PythonCHB (approved):
Really nice work!

Thanks for being such a great contributor to class.

I hope to see you next quarter.

PR #150, opened 2017-12-12
johnn lightning talk and class based mailroom (session 09)

johnnzz:
lightning talk material
* slides for lightning talk
* Ansible example plays for lightning talk

class based mailroom
* mailroom converted to use two classes, Donor() and Donors()
  note the emphasis was on the classes, so things like printing were not really 
  cleaned up.
* unit tests updated


PythonCHB:
Whoa! that's a lot of files!

I'll look at mailroom OO -- let me know if you want anything else reviewed.

students/johnn/session09/mailroom.py
```diff
@@ -0,0 +1,688 @@
...
+    a match(search) utility function.
+
+    get_donor can be used to return a Donor() object given the
+    donor id.
+    """
+
+    def __init__(self, donor=None):
+        self._donors = {}
+        # TODO:
+        #   donors should take a donor object or multiple donor objects
```
PythonCHB, line 32: good idea :-)

students/johnn/session09/mailroom.py
```diff
@@ -0,0 +1,688 @@
...
+        if donor:
+            self.add_donor(donor)
```
PythonCHB, line 34: good to call this method from here -- well done.

students/johnn/session09/mailroom.py
```diff
@@ -0,0 +1,688 @@
...
+
+    def __repr__(self):
```
PythonCHB, line 36: nice!

students/johnn/session09/mailroom.py
```diff
@@ -0,0 +1,688 @@
...
+        repr_out=[]
+        for key in self._donors:
+            repr_out.append( repr(self._donors[key]) )
+        return "Donors( " + ", ".join(repr_out) + " )"
+
+    def __str__(self):
```
PythonCHB, line 42: you can actually do a direct:

__str__ = __repr__

to make them the same function. Also, if you leave out the __str__, I think the __repr__just gets used anyway.

students/johnn/session09/mailroom.py
```diff
@@ -0,0 +1,688 @@
...
+        return self.__repr__()
+
+    def __iter__(self):
```
PythonCHB, line 45: very slick!

students/johnn/session09/mailroom.py
```diff
@@ -0,0 +1,688 @@
...
+
+    @property
+    def full_name_index(self):
+        """ Provide an index of keys, full_name, last_name and first_name """
+        index=[]
+        for key in self._donors:
+            index.append( (self._donors[key].full_name, key, 
+                self._donors[key].last_name,
+                self._donors[key].first_name ) )
+        return sorted(index)
```
PythonCHB, line 74: if you have it in a list, and want it sorted, it's a tiny bit more efficient ot sort the list, and then return it:

```
index.sort()
return index
```
rather than making a new list with sorted()

students/johnn/session09/mailroom.py
```diff
@@ -0,0 +1,688 @@
...
+    def match_donor(self, query_full_name):
+        """ 
+        find all records that match the query string
+
+        return tupple with full_name, id
+
+        """
+        matches=[]
+        for name, id, _, _ in self.full_name_index:
+            if query_full_name.strip().lower() == name.lower():
```
PythonCHB, line 85: probably a good idea to have one method that "normailizes" the name, and then use that method to build the full_name_index, and clean up the query_name.

but if you are doing an exact match like this, you could have the full_name_index be a dict, for fast lookup.

students/johnn/session09/mailroom.py
```diff
@@ -0,0 +1,688 @@
...
+    """
+
+    def __init__(self, 
+            id="", created="",
+            full_name="",
+            first_name="", last_name="", middle_name="", suffix="",
+            donations=None, donation=None):
+
+        # normally no id is passed in, so we create one
+        if id == "":
```
PythonCHB, line 169: is there a reason to pass in an id? this can be prone to error -- i.e. duplicated ids (OK, pretty darn unlikely with uuid)

students/johnn/session09/mailroom.py
```diff
@@ -0,0 +1,688 @@
...
+            id = str(uuid.uuid1())
```
PythonCHB, line 170: you might want: `uuid.uuid().hex` here -- though they are almost the same.

students/johnn/session09/mailroom.py
```diff
@@ -0,0 +1,688 @@
...
+        In [638]: repr(d)
+        Out[638]: "Donor( id='7e93d724-de25-11e7-bee5-0800274b0a84', first_name='John', 
+            last_name='Adams', donations=[{'amount': 1000.0, 'date': '2017-12-11Z'}, 
+            {'amount': 1000.0, 'date': '2017-12-11Z'}], created='2017-12-11T03:43:47.686457Z' )"
+        In [639]:
+
+    """
+
+    def __init__(self, 
+            id="", created="",
```
PythonCHB, line 163: I'd probably use None as the default here -- "" does evaluate False, so almost as good,but still, it is a valid string in most contexts.

students/johnn/session09/mailroom.py
```diff
@@ -0,0 +1,688 @@
...
+        self._id = id
+
+        # normally, no timestamp is passed in, so we create one
+        if created == "":
+            created = datetime.datetime.utcnow().isoformat() + "Z"
+        # keep track of record creation
+        self.created = created
+
+        # test to see if any of the name components are set
+        sub_name_set = any(sub_name != "" for sub_name in (
```
PythonCHB, line 181: remember that non-empty strings are True, so this can be:

```
any((first_name,
     middle_name,
     last_name,
     suffix))
```

students/johnn/session09/mailroom.py
```diff
@@ -0,0 +1,688 @@
...
+            first_name,
+            middle_name,
+            last_name,
+            suffix))
+     
+        # we don't expect full_name and a subset to be set at the same time
+        if full_name != "" and sub_name_set:
```
PythonCHB, line 188: pretty slick!

students/johnn/session09/mailroom.py
```diff
@@ -0,0 +1,688 @@
...
+            raise ValueError("You cannot set 'full_name' and a subset of the name at the same time.")
+
+        # set the name via full_name or via the constituent parts
+        if full_name != "":
```
PythonCHB, line 192: you are checking for full_name twice here, you could put the check for sub_name-set inside this test.

And you can just do `if full_name` -- empty strings are False

students/johnn/session09/mailroom.py
```diff
@@ -0,0 +1,688 @@
...
+            raise ValueError("You cannot set 'donations' and 'donation' at the same time.")
+
+        if donation != None:
+            # if a donation is passed in, initialize the donations, then add the donation
+            self._donations = list()
+            self.add_donation(donation)
+        else:
+            if donations != None:
+                # TODO donations need more validation
+                self._donations = donations
```
PythonCHB, line 210: at the very least, you want to wrap a list() around it -- so you get a new list, rather than a reference to the one passed in.

students/johnn/session09/mailroom.py
```diff
@@ -0,0 +1,688 @@
...
+            else:
+                self._donations = list()
+            
+    @property
+    def id(self):
+        """ return the donor id """
+        return self._id
+
+    @id.setter
```
PythonCHB, line 219: If you don't define a setter, you'll get an AttributeError anyway.

Though kinda nice to provide a nice message.

students/johnn/session09/mailroom.py
```diff
@@ -0,0 +1,688 @@
...
+    def id(self, value):
+        """ donor id cannot be changed once established """
+        raise AttributeError("You cannot change the 'id' after the record has been created.")
+
+    @property
+    def donations(self):
+        """ print all donations """
```
PythonCHB, line 226: print ??

students/johnn/session09/mailroom.py
```diff
@@ -0,0 +1,688 @@
...
+        return self._donations
+
+    @donations.setter
+    def donations(self, value):
+        """ allow bulk setting of donation entries """
+        self._donations = value
```
PythonCHB, line 232: you probably want to wrap a list() around it -- so you'll get a copy, and be assured that it's a list, rather than, say, a tuple:

```
self._donations = list(value)
```
and better to use a meaningful name, like, day "donations"

students/johnn/session09/mailroom.py
```diff
@@ -0,0 +1,688 @@
...
+
+    def add_donation(self, value):
+        """ add a single donation """
+        today = datetime.datetime.utcnow().date().isoformat() + "Z"
+        try:
+            value = float(value)
+        except ValueError:
+            raise ValueError("Donations must be values.")
+        self._donations.append( { "amount": value, "date": today } )
```
PythonCHB, line 241: this is a tricky one -- it's nice to have the structure, but dicts are pretty heavyweight. I might use a simple tuple:

(amount, date)

to keep it lighter weight.

PythonCHB: or keep a list of donation and a separate list of dates -- it's all in this class, so you can keep them in sync, but it lets you do things like sum(donations) a lot more easily.

students/johnn/session09/mailroom.py
```diff
@@ -0,0 +1,688 @@
...
+
+    @property
+    def number_donations(self):
+        """ return the number of donations """
+        return len(self._donations)
+
+    @property
+    def total_donations(self):
+        """ return a sum of the donations """
+        total_donations = 0
```
PythonCHB, line 251: sum and comprehensions:

`sum(d['amount'] for d in self.donations)`

students/johnn/session09/mailroom.py
```diff
@@ -0,0 +1,688 @@
...
+        for donation in self.donations:
+            total_donations += donation["amount"]
+        return round(total_donations, 2)
```
PythonCHB, line 254: I'm not sure rounding helps here -- you still get a float with all the digits. I usually wait 'till printing to round.

students/johnn/session09/mailroom.py
```diff
@@ -0,0 +1,688 @@
...
+
+    @property
+    def last_name(self):
+        """ return the last name """
+        return self._last_name
+
+    @last_name.setter
+    def last_name(self, value):
+        """ set the last name """
+        self._last_name = value.title()
```
PythonCHB, line 292: I'm not sure I'd bother with all this -- but it is a pretty nifty way to normalize the names.

I can't help thinking there would be a way to avoid the duplication of code here -- try to remember to bring this up when we talk about metaprogramming next quarter.

students/johnn/session09/mailroom.py
```diff
@@ -0,0 +1,688 @@
...
+        return " ".join(filter(None, [self.first_name, self.middle_name, self.last_name])) + suffix
+
+    @full_name.setter
+    def full_name(self, full_name):
+
+        """ allow the donor information to be updated via full_name as well """
+
+        # for simplicity's sake, we make the assumption a suffix will follow a comma
+
+        # capture suffix, if present
```
PythonCHB, line 330: you might want to put all this logic somewhere else -- in a static method, or a stand alone function in the module. It would make it easier to test and update.

But this is pretty slick!

students/johnn/session09/mailroom.py
```diff
@@ -0,0 +1,688 @@
...
+        suffix=""
```
PythonCHB, line 331: this could be simplified a bit by doing the split first (and use rsplit, maybe):

```
In [22]: "Roy Wood, Jr".rsplit(",", 1)
Out[22]: ['Roy Wood', ' Jr']

In [23]: "Roy Wood,Jr".rsplit(",", 1)
Out[23]: ['Roy Wood', 'Jr']

In [24]: "Roy Wood".rsplit(",", 1)
Out[24]: ['Roy Wood']

In [25]: "Roy H. Wood, Jr".rsplit(",", 1)
Out[25]: ['Roy H. Wood', ' Jr']
```

students/johnn/session09/mailroom.py
```diff
@@ -0,0 +1,688 @@
...
+    def _query_attributes(self, which_attributes=[], return_empty=True):
+        """ return list of formatted attribute/value entries """
+        attributes = []
+        for attr in which_attributes:
+            # for our purposes, never print internal attributes
+            if not attr.startswith("_"):
+                try:
+                    str_val=repr(getattr(self,attr))
+                    # don't return empty values if they don't want them
+                    if eval(str_val) or return_empty:
```
PythonCHB, line 374: I think this could be better done with some metaprogramming as well. So bring it up next quarter.

but to start -- a class instance  has a __dict__ attribute with all it's attributes in it -- you could probably make use of that.

Though I'm still not sure what you're trying to accomplish here...

students/johnn/session09/mailroom.py
```diff
@@ -0,0 +1,688 @@
...
+        """ return only the (settable) attributes needed to create the record """
+        which_attributes=["id", "first_name","middle_name","last_name","suffix",
+            "donations", "created"]
+        attributes = self._query_attributes(which_attributes, return_empty=False)
+        return "Donor( " + ", ".join(attributes) + " )"
+
+    def __str__(self):
+        """ return all the non-internal attributes """
+        which_attributes=[]
+        for attr in dir(self):
```
PythonCHB, line 391: you can also do: ``vars(self)`` which gives you the full set of attributes and values.

PythonCHB (approved):
Really nice work -- despite all my comments, they are all just tweaking style -- you've got great structure and  making really nice use of Pythonic features.

Above and beyond.

students/johnn/session09/mailroom.py
johnnzz: This was intended as stub functionality for a more fuzzy lookup, and I'm also trying to assume we would have valid cases where the names are exact, so I've been avoiding dicts.  But it's a good point in any case that this probably could be streamlined a bit.

students/johnn/session09/mailroom.py
johnnzz: The ability to pass in an ID was to make repr -> eval work.  Otherwise the object won't be "equal".

PR #153, opened 2017-12-16
Added challenge() to mailroom

Eowyn42:
ffmailroom and test_ffmailroom are working with the challenge() function to project amount needed to match donations between min and max. Marie helped me in office hours, at which time we realized that my set up doesn't actually need to use map() so the mult_donor function is written/tested but unused. 

Over break I intend to extend Chris's code in a way that DOES use map(), and frankly does a much better job with encapsulation by working more with objects. But it was faster to get my code extended before the Sunday deadline so that's what this pull request represents.

### DIFF
--- a/students/eowyn/session10/ff-mailroom/ffmailroom.py
+++ b/students/eowyn/session10/ff-mailroom/ffmailroom.py
@@ -1,0 +1,310 @@
+#!/usr/bin/env python
+import sys
+import copy
+from textwrap import dedent
+
+
+class Donor():
+    # Keep track of individual donors and their history of donations
+
+    def __init__(self, name):
+        self.donations = []
+        self.name = ''.join(name)  # should we check input?
+
+    def add_donation(self, amt):
+        self.donations.append(float(amt))
+
+    def generate_letter(self):
+        '''
+        Generate a formatted thank you note to a donor and return
+        the string of the thank you note for printing or writing.
+        '''
+        fs = "Thank you, {0}, for your generosity and recent gift of ${1:.2f}."
+        return fs.format(self.name, self.most_recent_donation)
+
+    def mult_donations(self, factor):
+        self.donations = list(map(lambda x: x * factor, self.donations))
+
+    def filter_donations(self, min_donation, max_donation):
+        if min_donation > max_donation:
+            (min_donation, max_donation) = (max_donation, min_donation)
+        self.donations = list(filter(lambda x: x > min_donation and
+                                     x < max_donation, self.donations))
+
+    @property
+    def sum_donations(self):
+        return round(sum(self.donations), 2)
+
+    @property
+    def count_donations(self):
+        return len(self.donations)
+
+    @property
+    def most_recent_donation(self):
+        return round(self.donations[-1], 2)
+
+    @property
+    def average_donation(self):
+        return round(self.sum_donations / self.count_donations, 2)
+
+
+class Transactions():
+    # Collect list of Donor objects and generate reports
+
+    def __init__(self):
+        self.all_donors = []
+
+    def add_donor(self, name, amt):
+        this_donor = self.get_donor(name)
+        if this_donor:
+            this_donor.add_donation(amt)
+        else:
+            this_donor = Donor(name)
+            this_donor.add_donation(amt)
+            self.all_donors.append(this_donor)
+        return this_donor.generate_letter()
+
+    def get_donor(self, name):
+        for i in self.all_donors:
+            if i.name == name:
+                return i
+
+    def list_names(self):
+        return ("Donor Names:\n" +
+                '\n'.join([i.name for i in self.all_donors]))
+
+    def generate_report_data(self):
+        names = [i.name for i in self.all_donors]
+        totals = [i.sum_donations for i in self.all_donors]
+        counts = [i.count_donations for i in self.all_donors]
+        averages = [i.average_donation for i in self.all_donors]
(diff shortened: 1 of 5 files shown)


students/eowyn/session10/ff-mailroom/ffmailroom.py
```diff
@@ -0,0 +1,305 @@
...
+
+    def generate_letter(self):
+        '''
+        Generate a formatted thank you note to a donor and return
+        the string of the thank you note for printing or writing.
+        '''
+        fs = "Thank you, {0}, for your generosity and recent gift of ${1:.2f}."
+        return fs.format(self.name, self.most_recent_donation)
+
+    def mult_donations(self, factor):
```
PythonCHB, line 25: looks good.

students/eowyn/session10/ff-mailroom/ffmailroom.py
```diff
@@ -0,0 +1,305 @@
...
+        self.donations = list(map(lambda x: x * factor, self.donations))
+
+    def filter_donations(self, min_donation, max_donation):
+        if min_donation > max_donation:
+            (min_donation, max_donation) = (max_donation, min_donation)
+        self.donations = list(filter(lambda x: x > min_donation and
```
PythonCHB, line 31: nice!

students/eowyn/session10/ff-mailroom/ffmailroom.py
```diff
@@ -0,0 +1,305 @@
...
+                                     x < max_donation, self.donations))
+
+    @property
+    def sum_donations(self):
+        return round(sum(self.donations), 2)
```
PythonCHB, line 36: something to think about -- round() still returns a floating point number, so you still have all the precision (about 15 digits, even though it's rounded.

So ususally we wait for final ouput or printing to round the numbers to even cents.

or:
use integers cents as your unit
or:
use the Decimal data type

Also -- the only operation that might create uneven cents is division (or multiplication by a non-integer...), so if your'e going to round, the avaging function should be the only place you need it.

students/eowyn/session10/ff-mailroom/ffmailroom.py
```diff
@@ -0,0 +1,305 @@
...
+        if this_donor:
+            this_donor.add_donation(amt)
+        else:
+            this_donor = Donor(name)
+            this_donor.add_donation(amt)
+            self.all_donors.append(this_donor)
+        return this_donor.generate_letter()
+
+    def get_donor(self, name):
+        for i in self.all_donors:
```
PythonCHB, line 68: maybe a more meaningful name than "i" could be good -- not a big deal with this little code, but still more clear.

students/eowyn/session10/ff-mailroom/ffmailroom.py
```diff
@@ -0,0 +1,305 @@
...
+            if i.name == name:
+                return i
+
+    def list_names(self):
+        return ("Donor Names:\n" +
+                '\n'.join([i.name for i in self.all_donors]))
+
+    def generate_report_data(self):
```
PythonCHB, line 76: this is looping through all_donors a bunch of separate times -- not very efficient if that was pull from a DB or something. Could you refactor to only loop once?Maybe you wouldn't need the zip then either.

students/eowyn/session10/ff-mailroom/ffmailroom.py
```diff
@@ -0,0 +1,305 @@
...
+            body = body + fs2.format(*entry) + '\n'
+        return body
+
+    def print_donor_records(self):
+        ''' Pretty-print the donor records '''
+        print(self.setup_table())
+        print(self.setup_body())
+
+    @property
+    def total_donations(self):
```
PythonCHB, line 112: you can use sum() here too:

total = sum( (d.sum_donations for d in self.all_donors) )

students/eowyn/session10/ff-mailroom/ffmailroom.py
```diff
@@ -0,0 +1,305 @@
...
+        total = 0
+        for i in self.all_donors:
+            total += i.sum_donations
+        return total
+
+    def challenge(self, factor, min_donation=0,
+                  max_donation=float('inf')):
```
PythonCHB, line 119: nice!

students/eowyn/session10/ff-mailroom/ffmailroom.py
```diff
@@ -0,0 +1,305 @@
...
+        if self.total_donations == 0:
+            return 0
+        new_transactions = copy.deepcopy(self)
+        for i in new_transactions.all_donors:
+            i.filter_donations(min_donation, max_donation)
```
PythonCHB, line 124: perhap a little more "functional" to pass the results of teh filter right into the map, but this is good, too.

students/eowyn/session10/ff-mailroom/test_ffmailroom.py
```diff
@@ -0,0 +1,201 @@
...
+    d.add_donation(amt)
+    assert d.name == name
+    assert d.donations == [345]
+
+def test_name():
+    name = "Ada Lovelace"
+    amt = 345
+    d = mr.Donor(name)
+    d.add_donation(amt)
+    assert d.name == name
```
PythonCHB, line 20: you already tested this in test_donor_init --no need to do it again. (not that it hurts, either)

PythonCHB (approved):
Really nice work.

I look forward to seeing what you do next quarter!

-Chris

PR #155, opened 2018-01-10
labs for session 11

mattmaeda:
create an iterable with range-like functionality
create a series of generators


students/mattmaeda/session11/generator_solution.py
```diff
@@ -0,0 +1,53 @@
...
+
+def intsum():
+    last = 0
+    for i in range(1, MAX_ITERATION):
+        yield last
+        last += i
+
+
+
+def intsum2():
```
PythonCHB, line 11: we actually had this in the tests because we had two different solutions :-)

students/mattmaeda/session11/generator_solution.py
```diff
@@ -0,0 +1,53 @@
...
+def fib():
+    def fibonacci(n):
+        if n == 0:
+            return 0
+        elif n == 1:
+            return 1
+        else:
+            return fibonacci(n-1) + fibonacci(n-2)
+
+    for i in range(MAX_ITERATION):
```
PythonCHB, line 31: so this is going to go back and do the full iteration each time. 

Can you do it with it marching up the numbers without going back to the beginning each time?

students/mattmaeda/session11/generator_solution.py
```diff
@@ -0,0 +1,53 @@
...
+            return False
+        if n % 3 == 0:
+            return False
+        if n % 5 == 0:
+            return False
+        if n % 7 == 0:
+            return False
+        return True
+
+    for i in range(2, MAX_ITERATION):
```
PythonCHB, line 51: there are a lot oftricks to make calculating primes more efficient.

can you make this a bit better?

But you got the generator part right.

Also -- what about larger numbers?

students/mattmaeda/session11/iterator_1.py
```diff
@@ -0,0 +1,72 @@
...
+        return self
+    def __next__(self):
+        self.current += 1
+        if self.current < self.stop:
+            return self.current
+        else:
+            raise StopIteration
+
+
+class IterateMe_2(IterateMe_1):
```
PythonCHB, line 29: you're overriding almost everything, so might as well just start from scratch.

Perhaps the word "extend" confused it -- I might change it to be more functional...

students/mattmaeda/session11/iterator_1.py
```diff
@@ -0,0 +1,72 @@
...
+    """
+    def __init__(self, start, stop=None, step=1):
+        real_stop = None
+
+        if stop is None:
+            real_stop = start
+        else:
+            real_stop = stop
+
+        super(IterateMe_2, self).__init__(real_stop)
```
PythonCHB, line 40: since the __init__ signature is changed, I'd not bother calling super here, and just initialize everything you need.

students/mattmaeda/session11/test_generator.py
```diff
@@ -0,0 +1,78 @@
...
+    assert next(g) == 1
+    assert next(g) == 2
+    assert next(g) == 3
+    assert next(g) == 5
+    assert next(g) == 8
+    assert next(g) == 13
+    assert next(g) == 21
+
+
+def test_prime():
```
PythonCHB, line 67: did this test pass? it looked like your prime check only went to 7 -- so would not have returned the ones after 11 correctly -- though maybe you get to 51??

PythonCHB: Try all of them up to 100:

2, 3, 5, 7, 11, 13, 17, 19, 23, 29, 31, 37, 41, 43, 47, 53, 59, 61, 67, 71, 73, 79, 83, 89, 97

students/mattmaeda/session11/iterator_1.py
```diff
@@ -0,0 +1,72 @@
...
+
+        if stop is not None:
+            self.current = (step * -1) + start
+
+        self.step = step
+
```
PythonCHB, line 46: you haven't overridden __iter__, so this won't reset correctly if re-started....

-CHB

PythonCHB (approved):
You got the iteration protocol and yield down -- nice work!

PR #156, opened 2018-01-10
generator tests and samples

alinafe82:
Figured out how to exclude the extra fluff, created a branch (git branch to-add-later), then removed it from master branch, I will clean up the new branch with non-duplcated items.  Still working on my workflow, but thought it may be of help to others.


students/alinafe/session10/generator_solution.py
```diff
@@ -0,0 +1,86 @@
...
+        lnum += i
+
+
+n = intsum()
+
+for i in n:
+    print(i, end=' : ')
+
+
+def intsum2():
```
PythonCHB, line 22: I only had an intsum2 test, 'cause I had an alternate solution.... :-)

students/alinafe/session10/generator_solution.py
```diff
@@ -0,0 +1,86 @@
+# import mem_profile
```
PythonCHB, line 1: nice find!

students/alinafe/session10/generator_solution.py
```diff
@@ -0,0 +1,86 @@
...
+# print('Memory (After): {}Mb'.format(mem_profile.memory_usage_resource()))
+# print('Took {} seconds'.format(t2 -t1))
+
+def doubler():
+    """Doubler numbers generator"""
+    # nums = [1,2,3,4,5]
+    # for t in nums:
+    # yield (t * 2)
+
+    lnum, hinum = 1, 20
```
PythonCHB, line 49: you'd really want to pass these in to the generator function -- that's what makes it so cool -- youcan create custom  generators with a function call....

students/alinafe/session10/generator_solution.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+# Fibonacci sequence
+
+def fib():
+    """Fib numbers generator"""
+    a, b, n = 1, 1, 30
+    for s in range(n):
+        yield a
+        a, b = b, a + b
```
PythonCHB, line 73: nice!

students/alinafe/session10/generator_solution.py
```diff
@@ -0,0 +1,86 @@
...
+# Prime numbers
+
+def nprme():
+    for i in range(2,8):
+        for j in range(i*2, 50, i):
+            yield j
+            i += 1
+
+def prime():
+    for x in range(2,50):
```
PythonCHB, line 84: even numbers are not prime, so you can do range (3, 50, 2)

but better to pass in the 50, or let it go on forever....

students/alinafe/session10/generator_solution.py
```diff
@@ -0,0 +1,86 @@
...
+        if x not in nprme():
```
PythonCHB, line 85: so this will work, but it will compute the whole series each time -- pretty inefficient!

students/alinafe/session10/generator_solution.py
```diff
@@ -0,0 +1,86 @@
...
+    a, b, n = 1, 1, 30
+    for s in range(n):
+        yield a
+        a, b = b, a + b
+
+# Prime numbers
+
+def nprme():
+    for i in range(2,8):
+        for j in range(i*2, 50, i):
```
PythonCHB, line 79: what if you wanted to go past 50?

students/alinafe/session10/generator_solution.py
PythonCHB: well, not the whole list, but it will start at the beginning each time.
and the list of non-primes is larger than the list of primes....

students/alinafe/session10/generator_solution.py
PythonCHB: This would get to be a really big list....

PythonCHB (approved):
you got the iterator protocol down -- and yield. Nice work.

PR #157, opened 2018-01-14
Mailroom package; Generator labs

Eowyn42:
Please ignore session10/

This PR includes a package of Mailroom, and solutions to the labs done in class last Tuesday.


students/eowyn/session11/Mailroom/setup.py
```diff
@@ -0,0 +1,29 @@
+from setuptools import setup
+
+def get_version():
+    """
+    Reads the version string from the package __init__ and returns it
+    """
+    with open(os.path.join("capitalize", "__init__.py")) as init_file:
```
PythonCHB, line 7: you're going to need to update this with your package name :-)

students/eowyn/session11/generator_solution.py
```diff
@@ -0,0 +1,61 @@
...
+def intsum(start=0, step=1):
+    i = start
+    total = start
+    while True:
+        yield total
+        i += 1
+        total += i
+
+
+def doubler(start = 1, step = 1):
```
PythonCHB, line 28: step??

students/eowyn/session11/generator_solution.py
```diff
@@ -0,0 +1,61 @@
...
+        i = i * 2
+
+
+def fib(a = 0, b = 1):
+    x = a
+    y = b
+    yield y
+    while True:
+        x,y = y, x+y
+        yield y
```
PythonCHB, line 41: nice!

students/eowyn/session11/iterator_1.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+class IterateMe_2():
+    def __init__(self, start=0, stop=5, step=1):
+        self.current = start-step
+        self.stop = stop
+        self.step = step
+        self.start = start
+    def __iter__(self):
+        return IterateMe_2(self.start,self.stop,self.step)
```
PythonCHB, line 35: you usually want to return self -- that is, this same object.

you can reset self.current so it restarts.

PythonCHB (approved):
All looks good!

PR #158, opened 2018-01-15
Session11

johnnzz:
While the setup files will allow the installation of the "mailroom" modules via "pip install ." and they can be import and run from within the program, a number of things puzzle me:

1) How is it normally handled to set up a system path to allow execution of the "scripts"?

2) How are package data files typically handled (conventions about location as well as implementation - does the setup script define this location and make it available to the script, or does the script simply use a convention?)


PythonCHB:
These two sections in the mailroom packaging exercise should help

https://uwpce-pythoncert.github.io/PythonCertDevel/exercises/mailroom-pkg.html#making-the-top-level-script-runnable

https://uwpce-pythoncert.github.io/PythonCertDevel/exercises/mailroom-pkg.html#including-data-files

But let me know if that's not clear.

students/johnn/mailroom/bin/mailroom
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env python3
```
PythonCHB, line 1: looks good!

students/johnn/mailroom/setup.py
```diff
@@ -0,0 +1,14 @@
+from setuptools import setup
+
+setup(
+    name='mailroom',
+    version='1.0.0',
+    author='johnn',
+    author_email='jnav@uw.edu',
+    packages=['mailroom', 'tests'],
+    scripts=['bin/mailroom'],
```
PythonCHB, line 9: so this should have put your mailroom script on the PATH when you install it -- does that not work?

students/johnn/mailroom/mailroom/ui.py
```diff
@@ -0,0 +1,250 @@
+#!/usr/bin/env python3
+
+import sys
+from pprint import pprint
+from mailroom.donors import Donors, Donor, save_donor_file
```
PythonCHB, line 5: you might want to try "relative imports" here:

`from .donors import Donors, ...`

students/johnn/session11/iterator_1.py
```diff
@@ -0,0 +1,34 @@
...
+        # if there is only one parameter, it's stop
+        if len(args) == 1:
+            stop = args[0]
+        else:
+            # everything from here on out is optional, so get
+            # what's there, and use defaults if not
+            try:
+                start = args[0]
+            except IndexError:
+                pass
```
PythonCHB, line 14: do you want to allow zero arguments?

students/johnn/session11/iterator_1.py
```diff
@@ -0,0 +1,34 @@
...
+            try:
+                stop = args[1]
+            except IndexError:
+                pass
+            try:
+                step = args[2]
+            except IndexError:
+                pass
+        self.current = start - step
```
PythonCHB, line 23: this all looks good. nice logic!

students/johnn/session11/iterator_1.py
```diff
@@ -0,0 +1,34 @@
...
+        self.step = step
+        self.stop = stop
+        #print("start", start, "stop", stop, "step", step)
+    def __iter__(self):
+        return self
```
PythonCHB, line 28: this won't reset itself if you try iterating through it again -- try that with range() and see how it behaves.

PythonCHB (approved):
looking good

PR #159, opened 2018-01-15
package for mailroom (sans tests)

Ginsusamurai:
need to make/include tests but this works for the base package

### DIFF
--- a/students/morgan/mailprog/bin/mailroom.py
+++ b/students/morgan/mailprog/bin/mailroom.py
@@ -1,0 +1,18 @@
+#!/usr/bin/env python
+
+import sys
+from mailprog.ui import main_loop
+from mailprog.dataModels import Storage
+from mailprog.functions import read_donors
+
+
+
+if __name__ == "__main__":
+    infile = "mailprog/data/donors.txt"
+    donor_objects = []
+
+    DB = Storage()
+
+    read_donors(infile, DB, donor_objects)
+
+    main_loop(DB)
--- a/students/morgan/mailprog/mailprog/functions.py
+++ b/students/morgan/mailprog/mailprog/functions.py
@@ -1,0 +1,80 @@
+from mailprog.dataModels import Donor
+
+def list_donors(DB):
+    print('\n')
+    for x in DB.donors:
+        print(x.name)
+    print('\n')
+
+def donation_input(DB):
+    donor_name = input("Enter a donor's full name \n"
+        "> ")
+    donor_donation = float(input("Enter a donation amount \n> "))
+
+    DB.add_donation(donor_name, donor_donation)
+
+def print_report(DB):
+
+    DB.donors = sorted(DB.donors, key=lambda x: x.total, reverse=True)
+    
+    print("\n\n")
+    print("{msg1: <20}|{msg2:<13}|{msg3:<14}|{msg4:<12}".format(msg1='Donor Name',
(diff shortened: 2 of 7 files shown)


PythonCHB:
you only want to put __init__.py files in the "package" dirs -- that is, files with modules you want to import.

so not in a data dir, and not in the bin dir -- it does have python file(s), but they should be top-level scripts, so they will be run, but not imported.

students/morgan/mailprog/bin/mailroom.py
```diff
@@ -0,0 +1,18 @@
...
+
+import sys
+from mailprog.ui import main_loop
+from mailprog.dataModels import Storage
+from mailprog.functions import read_donors
+
+
+
+if __name__ == "__main__":
+    infile = "mailprog/data/donors.txt"
```
PythonCHB, line 11: so this is the trick -- you are going to need to find a way to find that file when the package is installed in an arbitrary place. I'll talk about that a bit in class.

students/morgan/mailprog/mailprog/functions.py
```diff
@@ -0,0 +1,80 @@
+from mailprog.dataModels import Donor
+
+def list_donors(DB):
```
PythonCHB, line 3: if these were all in a class (maybe the Storage class), then they would all have access to the DB.

students/morgan/mailprog/mailprog/functions.py
```diff
@@ -0,0 +1,80 @@
...
+        f.write('Dear {name:},\n'
+            'Thank you for your support. You have donated a total of ${total:.2f}.\n'
+            'Thank you,\n'
+            '-The Students-\n'.format(name=x.name, total=x.total))
+        f.close()
+        print('Letter created for {}'.format(x.name))
+    print('\n')
+
+
+def read_donors(infile,DB, donor_objects):
```
PythonCHB, line 46: this in particular should probably be part of the Storage class.

students/morgan/mailprog/mailprog/mailroom.py
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env python
```
PythonCHB, line 1: isn't this now in bin? it should only be in one place.

students/morgan/mailprog/mailprog/ui.py
```diff
@@ -0,0 +1,33 @@
+from mailprog.functions import list_donors, donation_input, print_report, email_all
+
+
+
+infile = "mailprog/data/donors.txt"
```
PythonCHB, line 5: this was specified in the startup script, yes?

students/morgan/mailprog/mailprog/ui.py
```diff
@@ -0,0 +1,33 @@
+from mailprog.functions import list_donors, donation_input, print_report, email_all
```
PythonCHB, line 1: this is where OO is kinda nice --  you'd only need to import the name of the class

students/morgan/mailprog/setup.py
```diff
@@ -0,0 +1,11 @@
+from setuptools import setup
+
+setup(
```
PythonCHB, line 3: this all looks good.

PythonCHB (approved):
All looking pretty good -- but you really do need tests!!

PR #161, opened 2018-01-16
Initial stab at creating a package / inclass iterator_1.py work

mattbriggs:


students/briggsm/session10/generator_solution.py
```diff
@@ -0,0 +1,59 @@
+# Sum of the integers:
+# keep adding the next integer
+# 0 + 1 + 2 + 3 + 4 + 5 + …
+# 0, 1, 3, 6, 10, 15 …..
+
+def intsum():
+    total = 0
+    for i in range(100):
+        if i < 1:
```
PythonCHB, line 9: I think you'd get the same result without special casing zero.

students/briggsm/session10/generator_solution.py
```diff
@@ -0,0 +1,59 @@
...
+            yield total
+
+# Doubler:
+# Each value is double the previous value:
+
+# 1, 2, 4, 8, 16, 32,
+
+def doubler():
+    for i in range(1,100):
+        if i == 1:
```
PythonCHB, line 22: you could start with zero instead of special casing -- 2**0 == 1

students/briggsm/session10/generator_solution.py
```diff
@@ -0,0 +1,59 @@
...
+
+# 1, 1, 2, 3, 5, 8, 13, 21, 34…
+
+def recur_fibo(n):
+    if n <= 1:
+        return n
+    else:
+        return(recur_fibo(n-1) + recur_fibo(n-2))
+
+def fib():
```
PythonCHB, line 39: this is less than ideal -- it will go back and re-compute all of them each time.

Can you do a non-recursive solution ?

students/briggsm/session10/generator_solution.py
```diff
@@ -0,0 +1,59 @@
...
+# Generate the prime numbers (numbers only divisible by them self and 1):
+
+# 2, 3, 5, 7, 11, 13, 17, 19, 23…
+
+def prime():
+    for num in range(2,500):
+        for i in range(2,num):
+            if (num % i) == 0:
+                break
+        else:
```
PythonCHB, line 54: nice use of else on a for loop!

students/briggsm/session10/generator_solution.py
```diff
@@ -0,0 +1,59 @@
...
+        yield recur_fibo(i)
+    
+
+# Prime numbers:
+# Generate the prime numbers (numbers only divisible by them self and 1):
+
+# 2, 3, 5, 7, 11, 13, 17, 19, 23…
+
+def prime():
+    for num in range(2,500):
```
PythonCHB, line 50: no need to have it terminate -- and if you do want it to terminate, then you can pass the number of primes in to the generator function.

PythonCHB (approved):
looking good, looks like you got the idea with generators.

PR #163, opened 2018-01-21
Generators Uploaded

danwoj:
Please review my generators work saved within my "Session11" folder.


students/danwoj/Session11/generators.py
```diff
@@ -0,0 +1,59 @@
+import math
+
+def soi(start, stop):
+    '''
+    This is my Sum of Integers generator function
+    '''
+    i = start
+    for j in range(stop):
+        yield i
+        if j == 0:
```
PythonCHB, line 10: isn't this `+= j + 1`? - so no need for the special case?

students/danwoj/Session11/generators.py
```diff
@@ -0,0 +1,59 @@
...
+            i += 1
+        else:
+            i += j+1
+
+def doubler(start, stop):
+    '''
+    This is my Doubler generator function
+    '''
+    if start == 0:
```
PythonCHB, line 19: nice!

students/danwoj/Session11/generators.py
```diff
@@ -0,0 +1,59 @@
...
+        i = i * 2
+
+def fib():
+    '''
+    This is my Fibonacci Sequence function
+    '''
+    i, j = 0, 1
+    while True:
+        yield i
+        i, j = j, i + j
```
PythonCHB, line 33: perfect.

students/danwoj/Session11/generators.py
```diff
@@ -0,0 +1,59 @@
...
+
+def primes(start=2):
+    '''
+    This is my Prine Number function
+    '''
+    value = start
+    while True:
+        isprime = True
+        for x in range(2, int(math.sqrt(value) + 1)):
```
PythonCHB, line 42: not the most efficient prime number calculator -- but you go the generator part perfect.

students/danwoj/Session11/generators.py
```diff
@@ -0,0 +1,59 @@
...
+            if not value % x: 
+                isprime = False
+                break
+        if isprime:
+            yield value
+        value += 1
+
+def ecount(start=2, step=2):
```
PythonCHB, line 50: nice!

students/danwoj/Session11/test_generators.py
```diff
@@ -0,0 +1,79 @@
...
+
+def test_soi():
+    '''
+    Tests the Sum of Integers function
+    '''
+    start_value = 0
+    stop_value = 10
+    x = soi(start_value, stop_value)
+    y = next(x)
+    for i in range(stop_value-1):
```
PythonCHB, line 13: you can probably simply do:

for i in x:

here -- x is an iterator, so it should pick up where you left off in the for loop.

As long as it will terminate! Which in this case it does.

students/danwoj/Session11/test_generators.py
```diff
@@ -0,0 +1,79 @@
...
+    '''
+    Tests the Fibonacci Numbers function
+    '''
+    x = fib()
+    for i in range(10):
+        y = next(x)
+    assert y == 34
+
+    a = fib()
+    for i in range(1):
```
PythonCHB, line 47: not much point in looping here. :-)

PythonCHB (approved):
This all looks great!

PR #164, opened 2018-01-22
Added decorator to Mailroom and Solved Trapz several ways

Eowyn42:
- I moved all my mailroom files into /session12 but ONLY changed authorized_users.py and added a @record_users decorator to transactions.py
- I created a single trapz.py to calculate integrals, and created 3 test_trapz* files to test ways of creating the input functions. I did closures first and liked them best, did the *args,**kwargs form second, and the functools.partial third
- The *args, **kwargs approach needed a pair of little functions in trapz.py to construct the curve function before calling trapz(), and I realize that its probably possible to make one of them "decorate" trapz instead.
- functools.partial has to have keyword args. I think it is because otherwise it interprets arguments as "x". Anyway that's the only downside I can see of functools.partial. I think using partial with lambda functions would be pretty slick, but I haven't implemented it here, yet.


students/eowyn/session12/Mailroom/mailroom/__init__.py
```diff
@@ -0,0 +1 @@
+__version__ = "0.1.0"
```
PythonCHB, line 1: maybe time to increment the version :-)

students/eowyn/session12/Mailroom/mailroom/authorized_users.py
```diff
@@ -0,0 +1,13 @@
+""" 
+handle authorized users for database modification
+"""
+USERS = []
+
+def record_user(func):
+    print("inside decorator")
+    def get_user(self, *args, **kwargs):
```
PythonCHB, line 8: if you aren't doing anything with self, you can just do *args, **kwargs. self will end up as the first arg.

students/eowyn/session12/Trapz/test_trapz.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+def get_sloped_line(m, B):
+    ''' sloped line with slope m, intercept B'''
+    def sloped_line(x):
+        return m * x + B
+    return sloped_line
+
+
+def get_quadratic_eqn(A, B, C):
```
PythonCHB, line 22: nice -- currying is kinda cool, yes?

students/eowyn/session12/Trapz/trapz.py
```diff
@@ -0,0 +1,48 @@
+def xaxis(a, b, n):
+    ''' Generate floating x-points from a to b, inclusive '''
+    interval = (b - a) / n
+    while a <= b:
+        yield a
+        a += interval
```
PythonCHB, line 6: you need to be careful here -- if you keep adding a not-exact small number together, the errors accumulate.

it's a bit more accurate to multiply the interval times the number of steps so far each time.

you won't notice for relatively small number of values, but it can matter for, say millions...

students/eowyn/session12/Trapz/trapz.py
```diff
@@ -0,0 +1,48 @@
...
+    :param a: the start point for the integration
+    :type a: a numeric value
+
+    :param b: the end point for the integration
+    :type b: a numeric value
+    """
+    n = 1000
+    deltax = (b - a) / n
+    endpointssum = (fun(a) + fun(b)) / 2.0
+    xpoints = [x for x in xaxis(a, b, n)]
```
PythonCHB, line 26: this can just be:

`list(xaxis(a,b,n))`

students/eowyn/session12/Trapz/trapz.py
```diff
@@ -0,0 +1,48 @@
...
+    ypoints = [fun(x) for x in xpoints[1:-1]]  # excludes end points
```
PythonCHB, line 27: you do need to make a list first if you want to slice like this, but maybe there's a way to call the generator directly?

students/eowyn/session12/Trapz/trapz.py
```diff
@@ -0,0 +1,48 @@
...
+    area = deltax * (endpointssum + sum(ypoints))
+
+    return area
+
+
+
+def get_curve(fun, *args, **kwargs):
```
PythonCHB, line 34: so this kind of doing a generic currying -- clever.

But not necessary :-)

students/eowyn/session12/Trapz/trapz.py
```diff
@@ -0,0 +1,48 @@
...
+    ''' Pass arguments to form of function to get the curve f(x) '''
+    def curve(x):
+        return fun(x, *args, **kwargs)
+    return curve
+
+
+def trapzargs(fun, a, b, *args, **kwargs):
+    ''' Construct f(x) and calculate area under curve between a & b '''
+    somefunc = get_curve(fun, *args, **kwargs)
```
PythonCHB, line 43: you could define your curve function right here....

PythonCHB: or, when you call the fun:
` [fun(x) for x in xpoints[1:-1]] `

pass *args, **kwargs in right there. Then no need for the currying.

PythonCHB (approved):
Nice!

you're getting it! 

But I think you had surrying on teh brain, so didn't see that you could just  pass *args on through... :-)

PythonCHB:
>functools.partial has to have keyword args. I think it is because otherwise it interprets arguments as "x".

right -- it's s generic tool, so it needs to know which arguments you are trying to "Lock in".

> Anyway that's the only downside I can see of functools.partial. I think using partial with lambda functions would be pretty slick

not sure how that would work -- the point of partial is that you have a function with multiple args that you are going to re-use -- so you kinda need a name for it.

PR #165, opened 2018-01-22
Integration

guillaumethomas:
Hello Chris, 

Please find my homework regarding integration
this is very general in term of integration and management of the sub domains.
I will try to work on the extra credit tonight. 

Thank you, 

G.

### DIFF
--- a/students/guillaume/Q2_session02/integration.py
+++ b/students/guillaume/Q2_session02/integration.py
@@ -1,5 +1,5 @@
 #!/usr/bin/env python
-from math import cos
+from math import exp, isclose
 from numpy import linspace
 from itertools import islice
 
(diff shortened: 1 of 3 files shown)


students/guillaume/Q2_session02/integration.py
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env python
-from math import cos
+from math import exp, isclose
 from numpy import linspace
```
PythonCHB: kind of cheating -- I was expecting to to write this :-)

students/guillaume/Q2_session02/integration.py
```diff
@@ -40,6 +40,77 @@ def sectors(*args):
...
 
 
+def trapz_p(func, a, b, precision=10, **func_args):
+    '''
+    This function uses a more generic approach to trapeziodal numerical
+    integration in order to allow non constant size of the sub integration
+    sectors hence it is less efficient but more versatile
+    '''
+    # Test order of a & b
+    if a > b:
```
PythonCHB, line 50: nice!

students/guillaume/Q2_session02/integration.py
```diff
@@ -40,6 +40,77 @@ def sectors(*args):
...
+        a, b = b, a
+
+    def loc_f(x):
+        nonlocal func_args
```
PythonCHB, line 54: you don't need the nonlocal -- you aren't re-binding the name.

students/guillaume/Q2_session02/integration.py
```diff
@@ -40,6 +40,77 @@ def sectors(*args):
     return values, bands
 
 
+def trapz_p(func, a, b, precision=10, **func_args):
```
PythonCHB: for full flexibility, you want `*args, **kwargs`

students/guillaume/Q2_session02/integration.py
```diff
@@ -40,6 +40,77 @@ def sectors(*args):
...
+    '''
+    This function uses a more generic approach to trapeziodal numerical
+    integration in order to allow non constant size of the sub integration
+    sectors hence it is less efficient but more versatile
+    '''
+    # Test order of a & b
+    if a > b:
+        a, b = b, a
+
+    def loc_f(x):
```
PythonCHB, line 53: I like the internal currying!

students/guillaume/Q2_session02/integration.py
```diff
@@ -40,6 +40,77 @@ def sectors(*args):
...
+    '''
+    Assuming that f meets Riemann criterions for integration
+    default size of an integration sub domain is (1/20 x |b - a|)
+    for a sub dommain [x1, x2], where x2 >x 1
+    Z = |f(x2) - f(x1)| will be compared to a precision value
+    if Z bigger than precision x2 size will be decrease until the inequation
+    is accurate.
+    There is a test to avoid that calculations would run for too long
+    '''
+    func, a, b, precision = args
```
PythonCHB, line 82: why `*args`, rather than putting `func, a, b, precision` in the function signature?

students/guillaume/Q2_session02/integration.py
```diff
@@ -40,6 +40,77 @@ def sectors(*args):
...
+
+    def prec(func, precision):
+
+        def alpha(x, step):
+            f_x = func(x), func(x + step)
+            results = x + step, f_x[1]
+            if (abs(f_x[1] - f_x[0]) <= precision) or step < precision / 100:
```
PythonCHB, line 89: are you sure this is the right criteria? If I'm reading right, this is measuring the magnitude of the change in f with the step. But if the function is linear, you can have a huge change (steep slope), and still get the exact answer.

So what you want is some measure of non-linearity -- how different is the function from an actual trapezoid?

you might check and see how different it is if you use two steps rather than one?

you also need to be careful about adding up too many really small numbers - you can get cancellation effects. Not a big deal with double precision floating point, but something to keep in mind.

PythonCHB (approved):
nice work!

clever numeric method, and you've got the key point of the exercise down -- passing around functions and closures and currying.

students/guillaume/Q2_session02/integration.py
guillaumethomas: I build this criteria with the mindset of testing it on functions with various continuous parts but different definitions for each interval, I did not want to overkill so the calculation is relatively simple. Indeed linear function with a significant slope would be calculated very inefficiently. 
A very simple test on numerical derivation could help

PythonCHB: > A very simple test on numerical derivation could help

that's a good idea :-)

PR #166, opened 2018-01-22
iterator, generator, decorator, package, mailroom tests and start work on trapezoidal rule

Ginsusamurai:
added tests to mailroom package, generator files, added decorator to mailroom in the database class. Did some work on the trapezoidal rule problem I need some more work to get it figured out. having issues with the math.


students/morgan/advanced02/generator_solution.py
```diff
@@ -0,0 +1,43 @@
...
+    a = 2
+    while True:
+        check = True
+        for x in range(2,int((math.sqrt(a)+1))):
+            if (a%x)==0:
+                check = False
+                break
+        # if check == 0:
+        #     pass
+        if check == True:
```
PythonCHB, line 35: this can be just "if check:".

but better yet, you can use the else: clause in a for loop -- it will run if the loop wasn't broken out of.

students/morgan/advanced02/tes_trapez.py
```diff
@@ -0,0 +1,22 @@
+import trapez
```
PythonCHB, line 1: you need to call it "test_something" in order for pytest to find it. though this is fine if you pass it in to pytest.

students/morgan/advanced02/tes_trapez.py
```diff
@@ -0,0 +1,22 @@
...
+import math
+from iterator_1 import IterateMe_1
```
PythonCHB, line 3: are you using these in the tests?

students/morgan/advanced02/tes_trapez.py
```diff
@@ -0,0 +1,22 @@
...
+    area = trapez.trapezoid(trapez.line,0,10)
+
+    assert area == 50
+
+def test_sin():
+    area = trapez.trapezoid(math.sin,0,10)
+    print(area, 'area')
+
+def test_floatrange():
+    floatrange = IterateMe_1(0,10,1)
```
PythonCHB, line 15: ahh! but you want one that gives floating point numbers, somethign like:

1.0, 1.1, 1.2, 1.3, .....

students/morgan/advanced02/tes_trapez.py
```diff
@@ -0,0 +1,22 @@
...
+    for x in floatrange:
+        temp = x
+    assert temp == 9
```
PythonCHB, line 18: x is going to have the last value anyway, no need for temp here.

but you really want to check more than the last number anyway.

students/morgan/advanced02/tes_trapez.py
```diff
@@ -0,0 +1,22 @@
...
+
+def test_slope_line():
+    area = trapez.trapezoid(trapez.slope_line,0,10)
+    print(area, "slopeline")
```
PythonCHB, line 22: you'll want to put and `asset False` here, so that you'll see the print when you run the tests.

students/morgan/advanced02/trapez.py
```diff
@@ -0,0 +1,44 @@
+import math
+from iterator_1 import IterateMe_1
+
+
+
+
+
+
+def trapezoid(func, start, end, **kwargs):
```
PythonCHB, line 9: ideally you have both *args and **kwargs, so any function signature will do.

students/morgan/advanced02/trapez.py
```diff
@@ -0,0 +1,44 @@
...
+    #     # print(delta_x * i)
+    #     total += func(start + (delta_x * i))
+    total += func(end) / 2
+    return total * delta_x
+
+def line(x):
+    return 5
+
+
+def slope_line(x):
```
PythonCHB, line 35: you need an m and a B here:

```
def slope_line(x, m, B):
    ...
```
then you can figure out how to pass the m and B through your trapz function.

students/morgan/mailprog/mailprog/mailprog/dataModels.py
```diff
@@ -0,0 +1,181 @@
...
+
+    def __init__(self, name , donations= None):
+        self.name = name
+        self.donations = donations
+        self.i = 0
+
+    def add_donation(self, money):
+        self.donations.append(money)
+
+    def __iter__(self):
```
PythonCHB, line 22: if you use a generator here (very slick!) then you don't need the `__next__` method. or `self.i`.

students/morgan/mailprog/mailprog/mailprog/dataModels.py
```diff
@@ -0,0 +1,181 @@
...
+        return self.donors
+
+    def list_donors(self):
+        temp = []
+        for x in self.donors:
+            temp.append(x.name)
+        return temp
+
+    def record_user(func):
```
PythonCHB, line 68: this should probably be called `_record_user`, so it's clear you don't want to call it outside this class.

students/morgan/mailprog/mailprog/mailprog/dataModels.py
```diff
@@ -0,0 +1,181 @@
...
+        print('record user is being called')
+
+        def func_wrapper(self, *args, **kwargs):
+
+            print('func wrapper called')
+            if self.user == "":
```
PythonCHB, line 74: `if not self.user` works.

students/morgan/mailprog/mailprog/mailprog/tests/test_model.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+
+
+
+"""Donor class tests"""
+def test_create_donor():
+    bob = dataModels.Donor("Robert Dunbar", 1)
+    assert bob.name == "Robert Dunbar"
+    assert bob.donations == 1
```
PythonCHB, line 14: careful here -- if there is only one donation, `.donations` should probably still be a list. -- or an iterable of some sort.

students/morgan/mailprog/mailprog/mailprog/tests/test_model.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+"""Storage class tests"""
+def test_add_donor():
+    DB = dataModels.Storage()
+    bob = dataModels.Donor("bobbert", [50])
+    timmy = dataModels.Donor("Tim Tam", [10])
+    DB.add_donor(bob)
+    DB.add_donor(timmy)
+    assert len(DB.donors) == 2
```
PythonCHB, line 50: you might want some kind of "num_donors" property -- in anticipation of whne donors might not be a simple list.

students/morgan/mailprog/mailprog/mailprog/tests/test_model.py
```diff
@@ -0,0 +1,76 @@
...
+    DB.add_donor(bob)
+    DB.add_donor(timmy)
+    blurb = DB.list_donors()
+    assert blurb == ['bobbert','Tim Tam']
+
+def test_donation_input():
+    DB = dataModels.Storage()
+    bob = dataModels.Donor("bobbert", [50])
+    DB.add_donor(bob)
+    DB.add_donation("BOBBERT", 10)
```
PythonCHB, line 65: nice!

PythonCHB (approved):
I didn't review the old stuff.

But looking good! looks like you've got decorators down pretty well.

PR #167, opened 2018-01-23
Stuck on Trapz; slopped straight line

danwoj:
I'm not sure why I'm being so dense but i'm stuck on this test. When I run the test, I assert an area of 8 but get around 199. I'm not sure what I'm missing but I'm sure it's silly. Any feedback would be appreciated.


students/danwoj/Session12/trapz.py
```diff
@@ -0,0 +1,11 @@
+def trapz(fun, a, b):
+    '''
+    Trapezoidal Function
+    '''
+    x_values = [i for i in range(1,101)]
```
PythonCHB: Here's your issue -- you've got x from 0 to 100. You want 100 values, but you want them to be evenly spaced from a to b, not 0 to 100!

danwoj: Got it. Works as desired now. Thanks for the nudge.

PythonCHB: if you commit and push the change, it will show up in this PR, and I can then merge it.

PR #169, opened 2018-01-24
trapezoidal rule lab

ea8606:
is there any reason to make this a oo program?  Wasn't sure what I would put in the class if I made a class for this.


PythonCHB:
No reason for OO here. In fact, this is more a functional paradigm exercise.

students/EricAdams/trapezoidal_rule/decorator_trapezoidal_rule.py
```diff
@@ -0,0 +1,32 @@
...
+
+import trapezoidal_rule as tr
+
+
+def area_under_function(func):
+    def trapz_area_under_curve(func, n, a, b, step):
+        '''
+        Calculate the area under a curve using the
+        trapezoidal rule.
+        :func = the function that defines the curve
```
PythonCHB, line 11: Is this a standard format for docstrings? I don' think I've seen it. Ideally you want to use a standardaized format so they can be nicely foramttted by doc tools.

I usually use the default sphinx format:

```
:param func: the function that defines the curve
:type func: a callable object, with the first argument as the varying axis
```
Bu there are others, like the "numpy" format:

https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt#docstring-standard

And there may be others -- but better to use a common standard in any case.

students/EricAdams/trapezoidal_rule/trapezoidal_rule.py
```diff
@@ -0,0 +1,83 @@
...
+    # return sum(sigma_list)
+    # use a genexp to calculate each element of sigma
+    result = 0
+    sigma_gen = (func(a + i * step) for i in range(1, n))
+    for i in sigma_gen:
+        result += i
+    return result
+
+
+def func_a_plus_func_b_div_2(func, a, b):
```
PythonCHB, line 35: not much point in breaking these out to different functions -- they are one or a few lines, aren't likely to change, and don't need to be called from more than one place.

But if you do break them up, then it might make sense to put them in a callable class to keep it all in one namespace.

Or, alternatively,  define these utility function inside the main trapz_area_under_curve function, so again, one namespace.

students/EricAdams/trapezoidal_rule/trapezoidal_rule.py
```diff
@@ -0,0 +1,83 @@
...
+    :b = numeric type
+    :step = the width of the trapezoid.
+    :step = float type
+    '''
+    # (b-a)/n
+    area1 = b_minus_a_div_n(n, a, b)
+    area2 = func_a_plus_func_b_div_2(func, a, b)
+    area3 = sigma(func, n, a, b)
+
+    area = area1 * (area2 + area3)
```
PythonCHB, line 72: but this is right.

students/EricAdams/trapezoidal_rule/trapezoidal_rule.py
```diff
@@ -0,0 +1,83 @@
...
+    :n = integer type
+    :a = the starting point on the x axis for calculations
+    :a = numeric type
+    :b = the end point on the x axis for calculations
+    :b = numeric type
+    :step = the width of the trapezoid.
+    :step = float type
+    '''
+    # (b-a)/n
+    area1 = b_minus_a_div_n(n, a, b)
```
PythonCHB, line 68: careful with naming -- this isn't an area -- it's a scale factor.

students/EricAdams/trapezoidal_rule/decorator_trapezoidal_rule.py
```diff
@@ -0,0 +1,32 @@
...
+        :b = numeric type
+        :step = the width of the trapezoid.
+        :step = float type
+        '''
+        # (b-a)/n
+        area1 = tr.b_minus_a_div_n(n, a, b)
+        area2 = tr.func_a_plus_func_n_div_2(func, n, a)
+        area3 = tr.sigma(func, n, a, b, step)
+
+        area = area1 + area2 + area3
```
PythonCHB, line 27: note that the math is wrong here -- but you got it right below.

students/EricAdams/trapezoidal_rule/decorator_trapezoidal_rule.py
```diff
@@ -0,0 +1,32 @@
+#! /usr/bin/env python
+
+import trapezoidal_rule as tr
+
+
+def area_under_function(func):
```
PythonCHB, line 6: I hadn't thought to make a decorator out of this -- interesting!

but you don't need to re-implement -- you could call `trapz_area_under_curve` from in the decorator.

students/EricAdams/trapezoidal_rule/test_trapezoidal_rule.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+def test_b_minus_a_div_n_2():
+    result = b_minus_a_div_n(4, 2, 4)
+    assert result == 0.5
+
+
+def test_trapz_area_under_curve():
+    result = trapz_area_under_curve(f, 1000, 0, 4)
+    assert result > 33
```
PythonCHB, line 41: you could be a bit more precise with the values for teh tests.

`math.isclose()` is handy for that.

students/EricAdams/trapezoidal_rule/test_trapezoidal_rule.py
```diff
@@ -0,0 +1,42 @@
...
+    result = b_minus_a_div_n(4, 4, 4)
+    assert result == 0
+
+
+def test_b_minus_a_div_n_2():
+    result = b_minus_a_div_n(4, 2, 4)
+    assert result == 0.5
+
+
+def test_trapz_area_under_curve():
```
PythonCHB, line 39: good idea to test with more than one function!

PythonCHB (approved):
nice work -- I like the decorator idea -- not sure how practical it is, but pretty cool!

PR #171, opened 2018-01-29
Timer and Error Raising Context Managers

Eowyn42:
Solutions to homework on context managers. I didn't get to the lambda-function assignment but hopefully can tackle it this week for submission next week.

### DIFF
--- a/students/eowyn/session13/raises.py
+++ b/students/eowyn/session13/raises.py
@@ -1,0 +1,24 @@
+#!/usr/bin/env python
+
+"""
+detect whether correct exception was thrown otherwise raise it
+"""
+
+## would like to throw specific error
+
+class raises():
+
+    def __init__(self, ErrType):
+        self.ErrType = ErrType
+        pass
+
+    def __enter__(self):
+        return self.ErrType
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        if exc_type is self.ErrType:
+            return True
+        else:
+            raise exc_type
+            assert False, "Wrong error thrown"
+
--- a/students/eowyn/session13/timer.py
+++ b/students/eowyn/session13/timer.py
@@ -1,0 +1,45 @@
+"""
+Record how long something takes to run. Print result and write to file,
+if a file object is provided. Otherwise print to stdout (screen). An
+optional name argument allows for the function being timed to be
+given a name which is logged.
+"""
+
+import sys
+import contextlib
+
+
+class Timer():
+    def __init__(self, fileobj=sys.stdout, name=""):
+        if name:
+            self.name = name
+        else:
+            self.name = "This code"
+        self.fileobj = fileobj
+
+    def __enter__(self):
+        from time import clock
+        self.starttime = clock()
+        return self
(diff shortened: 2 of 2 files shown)


students/eowyn/session13/raises.py
```diff
@@ -0,0 +1,24 @@
...
+detect whether correct exception was thrown otherwise raise it
+"""
+
+## would like to throw specific error
+
+class raises():
+
+    def __init__(self, ErrType):
+        self.ErrType = ErrType
+        pass
```
PythonCHB, line 13: no need for a pass here. pass if for when there is literally nothing in a method.

students/eowyn/session13/raises.py
```diff
@@ -0,0 +1,24 @@
...
+
+    def __enter__(self):
+        return self.ErrType
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        if exc_type is self.ErrType:
```
PythonCHB, line 19: extra credit -- make it so you can pass in more than one error type.

students/eowyn/session13/raises.py
```diff
@@ -0,0 +1,24 @@
...
+            return True
+        else:
+            raise exc_type
```
PythonCHB, line 22: does this work? one the raise happens, the assert won't run -- and you do want the assert to happen.

the test will still fail, doe to the error raised, but you want the assertion instead.

students/eowyn/session13/timer.py
```diff
@@ -0,0 +1,45 @@
...
+given a name which is logged.
+"""
+
+import sys
+import contextlib
+
+
+class Timer():
+    def __init__(self, fileobj=sys.stdout, name=""):
+        if name:
```
PythonCHB, line 14: strings are immutable, so you can just put "This code" in as the default.

students/eowyn/session13/timer.py
```diff
@@ -0,0 +1,45 @@
...
+            self.name = name
+        else:
+            self.name = "This code"
+        self.fileobj = fileobj
+
+    def __enter__(self):
+        from time import clock
```
PythonCHB, line 21: you want this import at the top of the file.

then it'll be  global and can be accessed everywhere.

students/eowyn/session13/timer.py
```diff
@@ -0,0 +1,45 @@
...
+        self.fileobj.write(result)
+
+
+"""
+Refactor as a generator
+"""
+
+
+@contextlib.contextmanager
+def timer(fileobj=sys.stdout):
```
PythonCHB, line 39: kinda slick, huh?

students/eowyn/session13/timer.py
```diff
@@ -0,0 +1,45 @@
...
+    from time import clock
+    starttime = clock()
+    yield object()
```
PythonCHB, line 42: you could just yield None, too. a raw object won't be useful anyway.

PythonCHB (approved):
Looking good -- you've got it.

PR #172, opened 2018-01-30
Updated Trapz and Timer Context Manager

danwoj:
I updated my latest Trapz function (working more so but still not right) and my Timer context manager which works.


students/danwoj/Session12/trapz.py
```diff
@@ -0,0 +1,24 @@
...
+    '''
+    Trapezoidal Function
+    '''
+
+    # The next four lines are generating 100 evenly distributed 
+    # x coordinates between points a and b
+    x_values = []
+    x_values.append(a+((b-a)/100))
+    for i in range(1,101):
+        x_values.append(x_values[i-1]+((b-a)/100))
```
PythonCHB, line 14: It's a little better to calculate the value at each point:

``a + i*(b-a)/100``

rather than adding them up as you go -- each value is not exact, so the error accumulate.

students/danwoj/Session12/trapz.py
```diff
@@ -0,0 +1,24 @@
...
+
+#    if hasattr(coef):
+#        text = 'Has coef attribute'
+#        return text
+
+    area1 = (b - a)/(2 * len(x_values))
```
PythonCHB, line 20: be a bit careful with naming -- this is not an area :-) it's a coefficient.

students/danwoj/Session12/trapz.py
```diff
@@ -0,0 +1,24 @@
...
+    area2 = fun(x_values[0], *args, **kwargs) + fun(x_values[-1], *args, **kwargs)
+    area3 = 2 * sum(fun(x, *args, **kwargs) for x in x_values[1:-1])
+    tot_area = area1 * (area2 + area3)
+    return tot_area
```
PythonCHB, line 24: this looks pretty good to me -- what's missing?

students/danwoj/Session13/test_timer.py
```diff
@@ -0,0 +1,10 @@
+import pytest, math
+
+from timer import Timer
+
+def test_timer():
+    with Timer() as t:
+        for i in range(90000):
+            i = i ** 20
+
+    assert math.isclose(t.interval, 0.05, rel_tol=0.00, abs_tol=0.1)
```
PythonCHB, line 10: I'm not sure you can count on the interval being consistent -- I don't think I"d test for that.

And yes, it's really hard to test this kind of thing.

students/danwoj/Session13/timer.py
```diff
@@ -0,0 +1,15 @@
+'''
+This is my attempt at a basic timing context manager
+'''
+
+import time
+
+class Timer:
```
PythonCHB, line 7: looks good!

PythonCHB (approved):
All looks good to me -- you've got the concepts down.

PR #174, opened 2018-01-30
mailroom decorated and test file

ea8606:
decorated mailroom.py and test file included

### DIFF
--- a/students/EricAdams/decorating_mailroom/decorated_mailroom.py
+++ b/students/EricAdams/decorating_mailroom/decorated_mailroom.py
@@ -1,0 +1,348 @@
+#!/usr/bin/env python
+"""
+This is an object oriented version.
+Added decorators for monitoring use
+"""
+
+import sys
+import math
+from textwrap import dedent
+# import functools
+import datetime
+
+import make_donors
+
+# Utility so we have data to test with, etc.
+
+
+def monitor(method):
+    '''
+    Create a decorator to monitor each time a method is used
+    Log the method name, arguments and date/time used in a file
+    named monitor_mailroom.txt
+    '''
+    # @functools.wraps(method)
+
+    def log(*args, **kwargs):
+        result = method(*args, **kwargs)
+        name = method.__name__
+        arg_str = ', '.join(repr(arg) for arg in args)
+        kwarg_str = ', '.join(repr(kwarg) for kwarg in kwargs)
+        with open('monitor_mailroom.txt', 'a') as fopen:
+            fopen.writelines('Called: method: {} args: {} kwargs: {} time: {}\n'.format(
+                name, arg_str, kwarg_str, datetime.datetime.now()))
+        return result
+    return log
+
+
+def get_sample_data():
+    """
+    returns a list of donor objects to use as sample data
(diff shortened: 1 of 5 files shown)


ea8606:
the generator version of the time context manager is raising an AttributeError: __enter__.  Can't seem to fix it

students/EricAdams/decorating_mailroom/decorated_mailroom.py
```diff
@@ -0,0 +1,348 @@
...
+    Log the method name, arguments and date/time used in a file
+    named monitor_mailroom.txt
+    '''
+    # @functools.wraps(method)
+
+    def log(*args, **kwargs):
+        result = method(*args, **kwargs)
+        name = method.__name__
+        arg_str = ', '.join(repr(arg) for arg in args)
+        kwarg_str = ', '.join(repr(kwarg) for kwarg in kwargs)
```
PythonCHB, line 30: very nice -- great use of generator expressions!

students/EricAdams/decorating_mailroom/decorated_mailroom.py
```diff
@@ -0,0 +1,348 @@
...
+        with open('monitor_mailroom.txt', 'a') as fopen:
```
PythonCHB, line 31: this is something to think about -- this re-opens the file each time -- could you open it once and keep a reference around for the duration of the program running?

students/EricAdams/timing_context_manager/generator_timer_context_manager_write_to_file_obj.py
```diff
@@ -0,0 +1,35 @@
...
+import time
+import sys
+from contextlib import contextmanager
+
+
+class Timer:
+    def __init__(self, file_object=sys.stdout):
+        self.file_object = file_object
+
+    @contextmanager
```
PythonCHB, line 15: so here's the trick -- the contextlib.contextmanager decorator is designed to work wrap a generator function all by itself, not as a method on a class.

I think if you simply pulled this out of the class you'd have it :-)

students/EricAdams/timing_context_manager/timer_context_manager.py
```diff
@@ -0,0 +1,21 @@
+#! /usr/bin/env python
+'''
+Create a simple timer context manager
+'''
+import time
+
+
+class Timer:
```
PythonCHB, line 8: this looks fine

students/EricAdams/timing_context_manager/timer_context_manager_write_to_file_obj.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+class Timer():
+
+    def __init__(self, file_object=sys.stdout):
+        self.file_object = file_object
+
+    def __enter__(self):
+        self.start = time.clock()
+        self.name = 'timer_writing_to_file'
```
PythonCHB, line 16: it'd be nice to have a way for the user to give it their own name...

But this all looks good otherwise.

PythonCHB (approved):
looks good!

PR #175, opened 2018-01-31
context manager homework

mattbriggs:


students/briggsm/session13/context_time.py
```diff
@@ -0,0 +1,20 @@
...
+
+    def __exit__(self, *args):
+        self.end = time.clock()
+        self.duration = self.start - self.end
+
+
+with Timer() as t:
+    for i in range(100000):
+        i = i **20
+    print("This code took {} to run.".format(t))
```
codedragon, line 20: So, this looks like it is just printing the start time. In order to get the duration, you have to have your print statement outside of the context manager, since duration isn't set until the exit. Which means, incidentally, that you are including the time to print in the duration. Additionally, in order to be able to query the object outside of the context manager, you will also have to return self (instead of self.start) in the __enter__ block. Then you should be able to print t.duration.

students/briggsm/session13/myopen.py
```diff
@@ -0,0 +1,17 @@
...
+        self.flags = flags
+
+    def __enter__(self):
+        f = open(self.name, self.flags)
+        return self.f
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.f.close()
+        return False
+
```
codedragon, line 15: Definitely good to be aware of the importance of the return value of __exit__! Note that if you have no return statement, the default is None, which is not True, so exceptions will be raised. But, like you, I prefer to be explicit.

codedragon (approved):
Almost there. :)

PR #176, opened 2018-02-04
Git clean up & mailroom mock testing

Eowyn42:
Hey, I had to do some git cleanup so there's quite a bit in here that should be entirely ignored. The work for this week is entirely in  students/eowyn/session12/Mailroom/mailroom/

I have some tests that won't run ... basically I went to office hours today and got the hang of doing testing with mocked input. At home, I tried to expand my test coverage of my ui.py but for the life of me could not figure out how to mock stdout properly, and so attempts to test main loops resulted in infinite-tests that never complete, and attempts to test functions that print stuff are also stymied. I'm just stuck and have to stop working on this today.


students/eowyn/session12/Mailroom/mailroom/test/test_ui.py
```diff
@@ -0,0 +1,71 @@
...
+class TestUI():
+
+    def test_collect_donor_valid(self, mocked_input, interactor):
+        mocked_input.side_effect = ["Tommy Vietor", "350"]
+        output = interactor.collect_donor_input()
+        assert output == ("Tommy Vietor", 350)
+
+    def test_collect_donor_invalid(self, mocked_input, interactor):
+        # this seems hacky but I don't have assertRaises() available
+        mocked_input.side_effect = ["Tommy Vietor", "pundit the dog"]
```
codedragon, line 25: Use pytest.raises context manager to test that an exception is raised. Are you sure you want to be checking for a raised exception here, though?

students/eowyn/session12/Mailroom/mailroom/test/test_ui.py
```diff
@@ -0,0 +1,71 @@
...
+    #     with mocked_input.assertRaises(SystemExit):
+    #         interactor.quit_code()
+
+    def test_return_to_menu(self, mocked_input, interactor):
+        assert interactor.return_to_menu() is True
+
+    def test_main_loop1(self, mocked_input, interactor):
+        interactor.mainloop()
+        assert interactor.run_interactive_loop.assert_called_once()
+
```
codedragon, line 66: What is this test suppose to be testing? This is an infinite loop, since you start the interactive_loop, and are patching input, in this case to do nothing.

students/eowyn/session12/Mailroom/mailroom/test/test_auth_user.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+    def test_auth_user_decorator(self, mocked_input, decorated_function):
+        decorated_function[0]()
+        mocked_input.assert_called_once()
+
+    def test_auth_user_called_with(self, mocked_input, decorated_function):
+        decorated_function[0]()
+        mocked_input.assert_called_once_with("Please enter user name:\n")
+
```
codedragon, line 30: You don't need both of these two tests. The second one is more complete, so I would just use that one.

codedragon (approved):
This is great! I'm not sure what you meant about the print statements. Were there tests you wanted to do but didn't try? I didn't check your coverage. As is, it seems like there are only two tests not working, and I hope now you will see why those two aren't working. I would consider adding docstrings to your tests, so you will understand later what you did and why. :)

PR #180, opened 2018-02-07
more tests

Eowyn42:


Eowyn42:
Relevant code here is backcast.py 
This is the quite rough innards of a project to combine 10-min time series of wind speed, mechanical specs on power generation at different wind bins, and hourly electricity pricing time series to calculate gross revenue. Then determine the gross revenue for peak- and off-peak hours during 12 months of the year. It involves working with pandas dataframes and time series packages. It's flexible to allow the user to specify which hours are peak (even non-continuous hours). 

My next steps are to organize this into classes and make a package, and write tests for all of it. I'd really like to wrap the whole thing in a mailroom-style UI. 

Since it is at a rough stage any feedback is welcome. Perhaps you see a sensible way to split it up -- I'm considering a class for parsing the met data, at least once class for combining the data from different sources which involves handling their separate start/end times, looking for overlap, calculating a typical year, dealing with leap years, and handling different time steps. Another class to handle the peak/off peak calculations. Maybe a little class to read the power data. 

Thanks! We can use this example in class if you want. 

Edit: Gave it more thought and discussion -- my approach to move forward will be to do test-driven development of a UI that incrementally implements more and more of the underlying functionality, adding tests for it as I go along, and letting the development of the UI + tests help guide the separation of functionality into classes.  I've created a package for the UI with methods that "pass" pending implementation, which I intend to do in a test-driven way using the methodology developed for backcast.py

PythonCHB:
This sounds like a great plan!

Now to take a look -- sorry it's taken me so long.

students/eowyn/backcast.py
```diff
@@ -0,0 +1,252 @@
+import pandas as pd
+import numpy as np
+from tabulate import tabulate
+
+fname = 'M426_Processed_Data.txt'
```
PythonCHB, line 5: You so want to get this in a package sooner than later ;-) Or at least broken out into multiple files all in one dir, etc. (though might as well do the package).

But even before then, this stuff should be in a function at least -- mayabe "main()" (or "startup", or "initilize", or??)

students/eowyn/backcast.py
```diff
@@ -0,0 +1,252 @@
...
+of fields like "St Dev" that need to be combined to "StDev" and
+retaining only the numeric component of the sensor height (in m or ft)
+"""
+colnames = list(metdf)
+sensorInfo = []
+for col in colnames:
+    sensor = col.split(' ')
+    unitIndex, conversion = sensor[0].lower().find('m'), 1.0
+    if unitIndex == -1:
+        unitIndex, conversion = sensor[0].lower().find('ft'), 0.3048
```
PythonCHB, line 24: ouch! semi-ugly dataset!

probably not worth it for this but there are unit packages for python. none are perfect, but "pint" is pretty good.

https://pint.readthedocs.io/en/latest/

students/eowyn/backcast.py
```diff
@@ -0,0 +1,252 @@
...
+heights = [item[0] for item in sensorInfo]
+meanWindHeights = [a * b for a, b in zip(windAve, heights)]
+mxInd = meanWindHeights.index(max(meanWindHeights))
+
+# Drop all other columns from dataframe
+dropcols = colnames[:mxInd] + colnames[mxInd + 1:]
+metdf.drop(dropcols, axis=1, inplace=True)
+
+# Replace invalid data with nan
+metdf.iloc[:, 0] = metdf.iloc[:, 0].replace(-99.99, np.nan)
```
PythonCHB, line 41: I'm kind of surprised that's not something built in to Pandas somehow. But I guess not.

students/eowyn/backcast.py
```diff
@@ -0,0 +1,252 @@
...
+        return x
+    xbar = np.mean(x) # Mean, ignoring NA
+    xsd = np.std(x) # Standard deivation, ignoring NA
+    LL = xbar - 2*xsd # Lower limit for outlier detection
+    UL = xbar + 2*xsd # Upper limit for outlier detection
+    return x.map(lambda y: y if y > LL and y < UL else np.nan) # Change outliers to NA
+
+# Given a pandas series x, replace any NA with median of non-NA values
+def replace_na_with_median(x):
+    return x.fillna(np.mean(x))
```
PythonCHB, line 61: You say median, but are using mean -- careful!

students/eowyn/backcast.py
```diff
@@ -0,0 +1,252 @@
...
+print('Off Peak Hours Monthly Sum')
+result = offPeakDF.groupby(offPeakDF.index.month).agg({windname: np.mean, "Generation": sum, "Revenue":sum})
+result.index.names=["Month"]
+print(tabulate(result,headers = 'keys'))
+result.to_csv('OffPeakHours-Revenue.txt', sep='\t')
+
+
+
+
+
```
PythonCHB, line 252: you probably want to collect all the function in one (or more) modules, and then have a main script (or main function) that  runs through the procedure and calls those functions.

And probably group them into "clean up the data" and "do the actual analysis"

students/eowyn/windrevenue/windrevenue/__init__.py
```diff
@@ -0,0 +1 @@
+__version__ = "0.1.0"
```
PythonCHB, line 1: here we go! looking good!

students/eowyn/windrevenue/windrevenue/parse_met_data.py
```diff
@@ -0,0 +1,43 @@
...
+
+import pandas as pd
+import numpy as np
+
+class parse_met_data():
+
+    #  Given a pandas series, return series with outliers replaced by nan
+    #  For series with few unique variables, do nothing and return the series
+    #  The hepatitis dataset has many columns with 2 or 3 unique variables
+    def replace_outliers_with_na(x):
```
PythonCHB, line 22: no self? is this supposed to be a @staticmethod ?

And if so, maybe a stand alone function is fine.

students/eowyn/windrevenue/windrevenue/power_curve_tool.py
```diff
@@ -0,0 +1,24 @@
...
+Keep track of previously used power curves as .json files
+Print a list of available power curves
+Read new power curves
+Read and write power curves from/to .json files
+"""
+
+import pandas as pd
+import numpy as np
+
+class power_curve_tool():
```
PythonCHB, line 15: Without the content, I can't tell, but is there a good reason to put these in classes? If they are essentially stand alone functions, they can simple be functions in a module.

PythonCHB (approved):
Good start!

Issue #183, opened 2018-02-19
Bug in json_saveables Dict handling


johnnzz:
The following pull requests are prerequisites to reproducing the problem with my code:

- [PR 182](https://github.com/UWPCE-PythonCert/IntroPython-2017/pull/182) "bug in json_save list handling"
- [PR 184](https://github.com/UWPCE-PythonCert/IntroPython-2017/pull/184) "partial json_save support in johnn's mailroom"

My Donors() class employs a dict of dicts which appears not be be handled by the json_saveables code.  I'm trying to debug it, but so far haven't been able to get my head around it. 

The objects we are starting with:
```
In [4]: z
Out[4]: Donors( Donor( did='70e3d2b8-159e-11e8-b256-0800274b0a84', first_name='Jerico', last_name='Jonez', created='2018-02-19T17:58:07.779796Z' ) )

In [5]:
```

When converting the Donors() object into a dict, it looks good to me (yes, Donors() is a dict, with a uuid key):

```
{'__obj_type': 'Donors',
 '_donors': {'ada5118e-159a-11e8-b256-0800274b0a84': {'__obj_type': 'Donor',
   '_did': 'ada5118e-159a-11e8-b256-0800274b0a84',
   '_donations': [],
   '_first_name': 'Jerico',
   '_last_name': 'Jonez',
   '_middle_name': '',
   '_suffix': '',
   'audit': [{'action': 'did',
     'args': ('4a311ba2-159a-11e8-b256-0800274b0a84',),
     'kwargs': {},
     'time': '2018-02-19T17:28:24.868073Z',
     'user': 'fred johnes'},
    {'action': 'first_name',
     'args': ('jerico',),
     'kwargs': {},
     'time': '2018-02-19T17:28:24.868225Z',
     'user': 'fred johnes'},
    {'action': 'middle_name',
     'args': ('',),
     'kwargs': {},
     'time': '2018-02-19T17:28:24.868241Z',
     'user': 'fred johnes'},
    {'action': 'last_name',
     'args': ('jonez',),
     'kwargs': {},
     'time': '2018-02-19T17:28:24.868255Z',
     'user': 'fred johnes'},
    {'action': 'suffix',
     'args': ('',),
     'kwargs': {},
     'time': '2018-02-19T17:28:24.868267Z',
     'user': 'fred johnes'},
    {'action': 'add_donor',
[...]

PR #182, opened 2018-02-19
Json save

johnnzz:
The logging portions of this pull request are unimportant.  Mostly I'm trying to get the fix for saveables upstream sooner rather than later to avoid a merge conflict.

-john

When processing a list back into Python the following error is observed:

```In [1]: from mailroom.donors import Donors, Donor

In [2]: from mailroom import security

In [3]: security.user = "fred johnes"

In [4]: x = Donor(full_name="jerico jonez")

In [5]: j = x.to_json_compat()

In [6]: x2 = Donor.from_json_dict(j)
DEBUG: {'time': '2018-02-19T06:08:55.009846Z', 'user': 'fred johnes', 'action': 'did', 'args': ('5d767144-153b-11e8-b256-0800274b0a84',), 'kwargs': {}}
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
<ipython-input-6-6a54b09ecca8> in <module>()
----> 1 x2 = Donor.from_json_dict(j)

/home/johnn/uwpython/IntroPython-2017/solutions/metaprogramming/json_save/json_save/json_save_meta.py in from_json_dict(cls, dic)
    109         obj = cls.__new__(cls)
    110         for attr, typ in cls._attrs_to_save.items():
--> 111             setattr(obj, attr, typ.to_python(dic[attr]))
    112         # make sure it gets initialized
    113         # obj.__init__()

/home/johnn/uwpython/IntroPython-2017/solutions/metaprogramming/json_save/json_save/saveables.py in to_python(val)
    149             print("DEBUG:", item)
    150             try:
--> 151                 obj_type = item["__obj_type"]
    152                 obj = Saveable.ALL_SAVEABLES[obj_type].from_json_dict(item)
    153                 new_list.append(obj)

KeyError: '__obj_type'

In [7]: 
```

This is in the following code of saveables.py:
```
    @staticmethod
    def to_python(val):
        """
        Convert an array to a list.

        Complicated because list may contain non-json-compatible objects
        """
        # try to reconstitute using the obj method
        new_list = []
        for item in val:
            try:
                obj_type = item["__obj_type"]
                obj = Saveable.ALL_SAVEABLES[obj_type].from_json_dict(item)
                new_list.append(obj)
            except TypeError:
                new_list.append(item)
        return new_list
```

The problem is the Exception type is a KeyError, not a TypeError.  It is unclear if there are situations where a TypeError is also expected, so I just added the KeyError.

With this change the routine works as expected.

```
[...]

PR #184, opened 2018-02-19
Partial json save support

johnnzz:
This is incomplete and Donor(s) doesn't work ([due to Issue 183?](https://github.com/UWPCE-PythonCert/IntroPython-2017/issues/183)) and has a dependency on [PR 182](https://github.com/UWPCE-PythonCert/IntroPython-2017/pull/182), but I wanted to file it to make it easier to observe issue 183.


PythonCHB:
good catch!

This is because if you have a dict in a list, then it won't give you the TypeError -- you can ask for a key -- but it will give you a keyError, 'cause the '__obj_type' key won't be there.

I added tests for this, and fixed the code in my version, so I'll close this PR. But good catch!

PR #186, opened 2018-02-25
Term 2 Project

mattbriggs:
This is my in progress project for Term 2.


students/briggsm/Term2_project/baselinter/baselinter/baselinter.py
```diff
@@ -1,14 +1,27 @@
...
-LOCATION = "D:\\UW\\Intro-to-Python\\IntroPython-2017\\students\\briggsm\\Term2_project\\baselinter\\baselinter\\" # os.path.dirname(os.path.realpath('__file__'))
 THISDATE = str(datetime.date.today()) # e:\MB\Better-Butter\Projects\BaseLinter
 LIST_of_SETS = []
 LINTER_NAME = ""
 
+# Functions
+
 
 def open_set_json(filename):
     '''Opens a JSON file and returns the a list of sets.'''
```
codedragon, line 27: Won't let me put a comment next to the code below this, but there is a good opportunity for a list comprehension here:
list_of_sets = [set(value) for value in in_data.values()]

students/briggsm/Term2_project/baselinter/baselinter/baselinter.py
```diff
@@ -79,20 +92,27 @@ def select_update(choices, wordlist, wordloc):
 Select ----------------------------\n
```
codedragon, line 92: Meh, this interface won't let me make comments on the older stuff. Above, in `return_members`, you have a break after a return. The break does not do anything, since if you enter that conditional, you have already returned. Also, the return None is unnecessary, since python returns None by default. 

Regarding the function I am actually interrupting, `select_update`: above the print statement seems to be all about figuring out which words you will be having the user select from, and below the print statement is all about interacting with the user. I would separate these into two functions, so that you can test that you are presenting the user with what you think you are, without having to print and interact with the user.

students/briggsm/Term2_project/baselinter/baselinter/baselinter.py
```diff
@@ -79,20 +92,27 @@ def select_update(choices, wordlist, wordloc):
...
+        except ValueError:
+            print ("Please type a valid option. \n")
+            choosing = True
 
 
 def load_linter():
     '''Load linters'''
     global LIST_of_SETS
     global LINTER_NAME
-    global LOCATION
```
codedragon, line 116: Got rid of a global, cool. ;-)

codedragon:
Hi Matt, 
Great project! 
A couple of big picture things:

Files with tests in them should start with the word test, so instead of baselinter_test.py test_baselinter.py. This way testrunners find the test files in addition to the tests.

I've been thinking about how to get rid of the last two globals. Both of your main functions use these variables, so you could just send them in and out, however you have a third function, quit, which wouldn't have the same signature. If you have quit do a sys.exit, then you wouldn't need to check a return variable, and you could do something like this (made a silly re-imagining, which I tested only a little):

```
# initialize 
list_of_sets = 'hello'
linter = 'goodbye'

def check_file(a, b):
    a = 'new'
    print('check', a, b)
    return a, b

def load_linter(a, b):
    b = 'spam'
    print('linter', a, b)
    return a, b

def exit_app(*_):
    sys.exit()

while True:
    sel = input("1, 2, or 3")
    linter, list_of_sets = chooser[sel][1](linter, list_of_sets)
```
Finally, regarding your punctuation problem. That is a fun one. I would run a check on each word, by using string.punctuation and lower, and then checking if there is a difference between the old word and the new word. I would probably do punctuation and capitalization separately (check capitalization after stripping punctuation). It is a bit painful, but one thing you have going for you is that I think you only care about punctuation at the beginning or end of the word, and capitalization only at the beginning. You will want to return the punctuation and the position, or in the case of capitalization, you could do a boolean, and then use this returned information to correct the replacement word capitalization and punctuation. Hope this helps. Feel free to bug me, if you have questions.

codedragon (approved):
Great work.

PR #187, opened 2018-02-28
mailroom app with basic logging

mattmaeda:


students/mattmaeda/mailroom_app/mailroom/cli.py
```diff
@@ -0,0 +1,91 @@
...
+def get_selection():
+    """ Display menu option and get user input """
+    logging.info("Get menu selections")
+    selection = input(MENU)
+    logging.info("User selected option %s", selection)
+    return selection.strip()
+
+
+def send_thank_you():
+    """ Record a donation and send thank you message """
```
codedragon, line 35: I would consider breaking this into two functions, get_donor_name and get_donation_amount, or something like that. Your function would look something like this then:
```
def send_thank_you():
    donor = get_donor_name() # this function would run the while loop, add the donor to the database if it isn't there and return the name
    amount = get_donation_amount() # this function would run the second while loop, get the donation amount
    donor.add_donation(amount) 
    print(DB.send_letter(donor))
```

students/mattmaeda/mailroom_app/mailroom/model.py
```diff
@@ -0,0 +1,200 @@
...
+class Donor(object):
+    """
+    Donor class responsible for handling donor information
+    """
+    _donorname = None
+
+    def __init__(self, donor_name, initial_donations=None):
+        logging.info("Set up donor '%s'", donor_name)
+        self.name = donor_name
+        if initial_donations is None:
```
codedragon, line 28: Preferred way of saying this is `if not initial_donations:`

students/mattmaeda/mailroom_app/mailroom/model.py
```diff
@@ -0,0 +1,200 @@
...
+
+    @property
+    def last_donation(self):
+        """ get the last donation made by donor """
+        logging.info("Get the last donation for %s", self.name)
+        try:
+            return self.donations[-1]
+        except IndexError:
+            logging.warn("Donor %s has not made any donations", self.name)
+            return None
```
codedragon, line 78: Python returns None by default.

students/mattmaeda/mailroom_app/mailroom/test/test_cli.py
```diff
@@ -0,0 +1,129 @@
...
+        assert option == "1"
+
+
+@mock.patch('sys.stdout', new_callable=StringIO)
+def test_send_thank_you(mock_stdout):
+    """ test send thank you """
+    user_input = ["Matt", "100"]
+    with mock.patch("builtins.input", side_effect=user_input):
+        cli.send_thank_you()
+        assert mock_stdout.getvalue() == THANK_YOU_OUTPUT
```
codedragon, line 65: Looks like you got sys.stdout patched! I think there were others having difficulties with this, so you might want to post a link to your tests in slack so others can see what you did.

codedragon (approved):
Logging, and tests and packaging, oh, my. Looks great. I didn't run the tests or the code, but everything looks good. I didn't see any configuration for logging, so you should probably add that (unless I just missed it somehow). Good to know how to change the defaults, even if you don't care about logging to a file in this particular case.

PR #188, opened 2018-03-06
Continued work on wind revenue project (code/peer review??)

Eowyn42:
IT IS FINISHED. 

This package, windrevenue, calculates the gross revenue of a single wind turbine given historical wind speed and power pricing data, and a power curve table to determine power generation from wind speed. Linear interpolation is used to calculate power generation at intermediate values, as power curve tables are on 0.5m/s step increments. 

I prioritized having a working UI that is able to demo the gross revenue calculation from start to finish, and having test coverage over the bulk of the core functionality. The design was definitely the hardest part of this project. There are 3 different kinds of data being ingested and data has to flow between the different classes in an orderly way. Encapsulation of concerns and isolation of data flows -- that is, passing data on a need-to-know basis and ensuring it flows downstream -- were guiding principles. 

In a very generalized sense, the UI instantiates each of 6 classes and uses those instances to encapsulate the data relevant to them. The underlying data structures are mostly pandas dataframes, which enables the backend to do a lot of time series manipulations.

In the future, I want to develop end-to-end tests, expand the UI to include more setting/getting/inspection for the data, and implement a power curve repository, such that the user could select from previously-used power curves.


students/eowyn/windrevenue/windrevenue/UI.py
```diff
@@ -27,7 +56,8 @@ def return_to_menu(self):
         ''' Return True to trigger exit out of sub-loop'''
         return True
 
-    def get_user_input(self, prompt_string):
+    @staticmethod
+    def get_user_input(prompt_string):
```
codedragon, line 60: I'm not sure this is buying you anything. I think it may be simpler and more readable to just put the try/except clause in the while loop.

students/eowyn/windrevenue/windrevenue/align_data.py
```diff
@@ -0,0 +1,139 @@
...
+    If there is 1 year of overlap, truncate data to that year
+    If not, calculate a typical year from all available data
+    Handle time axes at each step
+    Return a single data frame of 1 year of aligned hourly data
+    """
+
+    @staticmethod
+    def get_typical_year():
+        # Construct a dummy TimeStamp index for the typical year
+        return pd.date_range('2015-01-01', periods=8760, freq='H')
```
codedragon, line 28: This could be a property, AlignData.typical_year.

students/eowyn/windrevenue/windrevenue/align_data.py
```diff
@@ -0,0 +1,139 @@
...
+    def resample_timeseries(self, timestep='60min'):
+        """
+        Resample met, generation, and power pricing data to hourly
+        TO DO: Handle missing vals introduced by re-indexing better
+        """
+        repwr = self.pricing.get_pricing_field().resample(timestep).mean()
+        self.power_hour = repwr.bfill()
+        remet = self.met.get_wind_and_generation()
+        remet = remet.resample(timestep).mean()
+        self.met_hour = remet.bfill()
```
codedragon, line 43: I know it fills up __init__ with a bunch of variables, but it is a good practice to put all instance variables in the init, and just set them to None if they can't be set yet. See here for a good discussion: https://stackoverflow.com/questions/38377276/instance-variables-in-methods-outside-the-constructor-python-why-and-how

students/eowyn/windrevenue/windrevenue/electricity_pricing.py
```diff
@@ -1,24 +1,73 @@
...
-        pass
+    def load_new(self, fname=None):
+        """
+        Read pricing time series data from file, provided or default.
+        Store all pricing data as data frame, and select first
+        substation as the powerVar.
+        """
+        if fname is None:
+            from windrevenue.UI import UI
+            fname = UI.get_user_input("Please specify pricing file:\n")
```
codedragon, line 29: The UI should be controlling, not this class. Iow, the UI should be calling into this method, not the other way around. That way, you could stick a different front end (like a web api) on to your class, and you would not have to make any changes to the back end.

students/eowyn/windrevenue/windrevenue/electricity_pricing.py
```diff
@@ -1,24 +1,73 @@
...
+        Update self.windvar and return time series of chosen sensor.
+        """
+        print("SELECT FROM:\n")
+        colnames = list(self.powerdf)
+        self.print_substation_info()
+        choice = input("\n>>")
+        menu = dict(zip(range(len(colnames)), colnames))
+        if int(choice) in menu.keys():
+            self.powerVar = colnames[int(choice)]
+            print("Using substation: ", self.powerVar)
```
codedragon, line 73: This also looks like stuff that should be in the UI. Notice that this function has 2 purposes, one is prompting the user, the other updating data. It should only be updating and returning data.

students/eowyn/windrevenue/windrevenue/parse_met_data.py
```diff
@@ -1,43 +1,145 @@
...
+        self.pct = pct
+
+    def parse_met_file(self, fname=None):
+        """
+        Read met data file, store in data frame and store sensor choice
+        """
+        if fname is None:
+            from windrevenue.UI import UI
+            prompt_string = "Please specify the met time series file:\n"
+            fname = UI.get_user_input(prompt_string)
```
codedragon, line 37: You already know what I am going to say here. ;) I promise not to mention it again. :)

students/eowyn/windrevenue/windrevenue/peakhours.py
```diff
@@ -1,19 +1,54 @@
...
+        else:
+            self.off_peak_hours = pd.Series([i for i in range(12, 24)])  # 12..23
+
+    def print_peak_hours(self):
+        print("Peak hours: ", self.peak_hours)
+        print("Off-Peak hours: ", self.off_peak_hours)
 
     def get_peak_hours(self):
-        pass
+        return (self.peak_hours, self.off_peak_hours)
```
codedragon, line 35: This could also be a property.

students/eowyn/windrevenue/windrevenue/test/test_revenue.py
```diff
@@ -0,0 +1,171 @@
...
+    def test_add_revenue_value_spot_check(self, sample_data):
+        rev = GrossRevenue(aligned_data=sample_data)
+        rev.add_revenue_column()
+        z = rev.aligned_data.iloc[42, :]
+        assert isclose(z[0], 9.14, rel_tol=1e-04)
+        assert isclose(z[1], 2517.049924, rel_tol=1e-04)
+        assert isclose(z[2], 25.164100, rel_tol=1e-04)
+        assert isclose(z[3], 6.333930, rel_tol=1e-04)
+
+    def test_subset_data_peak(self, sample_data):
```
codedragon, line 71: This test looks like it should probably be broken up, at least in to two different tests (one for peak, one for off-peak). You can do your assert loops as: `assert oncount.count(4380) == len(oncount)` Alternatively you could use set, although count is faster in this instance.

codedragon (approved):
Great work! This is a large project, and you did a great job pulling it together. Some general comments: Nice job making small, testable functions. The general layout of code and tests is great. There is a clear separation of concerns, which doesn't happen often in first projects. Would be nice to see a README with an overview of the package, installation instructions, etc. Tried to use your ui, but got frustrated. I did later find some instructions in the UI file itself, which seemed to explain part of the problem. It appears you can’t use menu item (5) until you have loaded some data? Probably that shouldn’t be a menu option until after you have loaded data, then. Also, there was a reference to sample data being in sample_data/*txt, but I did not see this folder. But, these are fairly minor details.
Mostly, I would work at pulling all of the input into the UI. Congrats, you did an excellent job, and have clearly learned a lot of python so far.

Eowyn42:
Thank you very much for the detailed feedback! I’m on my way to Beijing and
will be back in a fortnight, at which time I’ll reach out and ask for a
grade equivalency letter similar to last quarter.

Best wishes,
Eowyn

PR #194, opened 2018-03-13
Class project

combatraffi:
Sorry to spam you with this.  Most recent updates should have the project.  I don't need you to review other work.


PythonCHB:
A little more description of the goal of the project would be nice.

students/Dkuchan/ClassProject/PythonClassProject.py
```diff
@@ -0,0 +1,625 @@
...
+teamlist2 = []  #HOLDS THE TEAM OBJECTS
+montelist = []
+teamdict = {}
+monteblue = [0, 0, 0]
+montered = [0, 0, 0]
+
+
+class Match:
+
+    # A match is a collection of data which represents a match FOR A SINGLE TEAM.
```
PythonCHB, line 29: not that I really understand the problem here, but it seems a "match" involved more than one team (2?) -- so you probably want a match object that captures that -- so you can't get things like "winner" out of sync.

students/Dkuchan/ClassProject/PythonClassProject.py
```diff
@@ -0,0 +1,625 @@
...
+        return 'Match {}, Team {}'.format(self.matchnumber, self.team)
+
+    #def __iter__(self):
+
+class MonteResult(Match):
+    # this class is intended to contain results for monte carlo runs
+    # it is a sublcass of Match as a monte run is just simulated Matches
+    # functions of match are not all used at this point.  perhaps with future updates
+
+    def __init__(self, bluescore, redscore, runnum):
```
PythonCHB, line 75: you haven't initialised the Match class -- so this won't have all the attributes of a Match. If this works, then this really isn't a subclass...

students/Dkuchan/ClassProject/PythonClassProject.py
PythonCHB: I would expect that the match object would have the teams involved in the match, and anything related to the match that isn't team-specific -- things like who won, the score, how long it took, or whatever.

students/Dkuchan/ClassProject/PythonClassProject.py
```diff
@@ -0,0 +1,625 @@
...
+
+   # def __repr__(self):
+    #    return '{}, {}: {}'.format(self.team, self.matchnumber, self.win)
+
+    def __str__(self):
+        return 'Team Number: {}, Team Name: {}'.format(self.number, self.name)
+
+    def fillmatches(self):
+        for i in range(0, len(matchlist)):   # I wanted to do this in a list
+            if int(matchlist[i].team) == int(self.number):  # did your team play in this match
```
PythonCHB, line 134: shouldn't these already be ints?

and you are only using i to pull out an item form the matchlist -- so you could just loop through the matchlist.

I _think_ this listcomp will do:

self.matchesplayed = [ match.number for match in matchlist if match.team == self.number ]

students/Dkuchan/ClassProject/PythonClassProject.py
```diff
@@ -0,0 +1,625 @@
...
+                self.matchesplayed.append(matchlist[i].matchnumber)  # if so append the list comprehensions but I is a match type variable
+
+    def fillmatchwins(self):
+        for i in range(0, len(matchlist)):   # I wanted to do this in a list
```
PythonCHB, line 138: similar to above

students/Dkuchan/ClassProject/PythonClassProject.py
```diff
@@ -0,0 +1,625 @@
...
+
+    
+
+
+                    #self.matchwins = {matchlist[i].matchnumber: matchlist[i].win for i in matchlist if matchlist[i].teamnumber == self.number}        
+       # self.matchesplayed = [int(matchlist[i].matchnumber) for i in matchlist if int(matchlist[i].teamnumber) == self.number]
+
+
+    def fillgearsdelivered(self):
+        for i in range(0, len(matchlist)):   #I wanted to do this in a list
```
PythonCHB, line 200: looking at these loops, it seems maybe keeping the matchlist up to date when each match is created would make more sense. Maybe have an object that represents the tournament?

students/Dkuchan/ClassProject/PythonClassProject.py
```diff
@@ -0,0 +1,625 @@
...
+    # this function creates a comparison of 2 teams for quick reference.
+    print()
+    print("Welcome to the Team Comparison Tool")
+    print("Which teams would you like to compare?: ")
+    print()
+
+
+def populateteamdict():
+    # this function creates a dict where the key is
+    # the team number (string) and the value is the team object.
```
PythonCHB, line 297: I would use either a number or a string for a team id -- if it's a number, use an int. or just let it be an arbitrary string.

students/Dkuchan/ClassProject/PythonClassProject.py
```diff
@@ -0,0 +1,625 @@
...
+
+def createteamlist():
+    # This very Janky function creates a list of team numbers, a matching list of team objects and calls the teamdict function
+    # I really dont like this function it is not really necessary and poorly concieved but I could not get the "elegant" way
+    # to work right and I needed something that worked in this case.
+    for i in range(0,len(matchlist)):
+        testteamnum = matchlist[i].team
+        if testteamnum not in teamlist:
+            teamlist.append(testteamnum)
+    for j in range(0,len(teamlist)):    #I THINK THIS IS SUPER JANKY BUT I COULD NOT GET A DIRECT CLASS LIST CREATION TO WORK! :(
```
PythonCHB, line 325: you almost never want to do this.

```
for team in teamlist:
    teamlist2.append(Team(team))
```
or
```
teamlist2 = [Team(t for t in teamlist)]
```
But what kind of objects are stored in teamlist? why not the Team objects themselves?

students/Dkuchan/ClassProject/PythonClassProject.py
```diff
@@ -0,0 +1,625 @@
...
+        teamlist2.append(Team(teamlist[j]))
+    populateteamdict()
+    fillteamstats()
+
+def readmatchfile():
+    # this function imports a match file CSV
+    # in the future I would like this to be done in a database.
```
PythonCHB, line 332: good idea!

students/Dkuchan/ClassProject/PythonClassProject.py
```diff
@@ -0,0 +1,625 @@
...
+
+
+def produceinsights(teamnum):
+    # This function calls multiple functions to provide a report
+    # containing statistical outcomes that are applicable
+    # to the specified team.
+    pass
+
+
+def doesteamexist(team):
```
PythonCHB, line 373: could you make this simply:

`team in teamlist`

if the types are right -- and comparison of the Team objects works. (define `__eq__`)

students/Dkuchan/ClassProject/PythonClassProject.py
```diff
@@ -0,0 +1,625 @@
...
+    for i in range(0, len(teamlist)):
+        if team == str(teamlist[i]):
+            existance = True
+    return existance
+
+
+def duplicatecheck(team, array):
+    # takes a team and an array, checks if it occurs more than once
+    duplicatestat = False
+    if array.count(team) >= 1:
```
PythonCHB, line 385: if this works, then `is` should work :-)

students/Dkuchan/ClassProject/PythonClassProject.py
```diff
@@ -0,0 +1,625 @@
...
+        duplicatestat = True
+    return duplicatestat
+
+def executemonte(matchsetuparray, runs=100000, factor = .2, rangemin = -1, rangemax = 1):
+    # this functions executes a monte carlo run.
+    # it is nominally called from the setupmonteUI fnc.
+    # currently this monte carlo ONLY works on gear scores
+    grossresults = []   #this will be a memory hog try to think of how to use a generator
```
PythonCHB, line 393: good idea...

but I don't see where this is used....

students/Dkuchan/ClassProject/PythonClassProject.py
```diff
@@ -0,0 +1,625 @@
...
+
+        #montescores.append(MonteResult(bluescore, redscore, i)) #creates an instance of monte match.monte
+        #dump detailed result file.
+
+
+def showteamdata(teamnumber):
+    print()
+    print()
+    print(str(teamdict[teamnumber]))
+    print("Matches Played: {}".format(teamdict[teamnumber].nummatches))
```
PythonCHB, line 428: probably a goo didea to create a variable:

`team = teamdict[teamnumber]`

then use `team` everywhere else...

PythonCHB (approved):
looks good, nice work.

Do think about a data model that keeps a bit more of the match data, rather than having to look things up all the time. And maybe more dicts, rather than lists, so lookups can be fast.

Enjoy the next class -- maybe you can put a webUI on this.

PR #195, opened 2018-03-14
updated with working labyrith creation and partial tests

Topcat7:
Still working on putting everything into a package and finishing up the unit tests.

Any input is appreciated.


codedragon:
I downloaded and tried it out. Pretty cool, nice job. I went through the code, trying to find places to make comments, but didn't see much. Partly because I don't know pygame, but also because your code looks pretty clean. It is nice using a framework like pygame when you are starting, since it makes it easier to see how classes should be delineated. There are a couple of instance variables in your classes that would be better initiated in __init__,  rather than in the methods that set them, but that is minor.  There are probably more tests that could be written, like what happens when a player hits a wall, but overall looks great. How hard was it to figure out the random maze generation?

Topcat7:
Thank you very much for your feedback.  I agree completely that more tests could be written, I struggled a fair bit with those trying to figure out what to test and how within the pygame framework.  Truthfully it was pretty difficult for me to get the random maze generation to work correctly, I had to research a lot about the way maze algorithm worked and how that would apply to my situation.  In the end though I was happy with the way it turned out.

PR #196, opened 2018-03-16
Final project

guillaumethomas:
Maria, Chris,

This is my ongoing work project.
This is an API Wrapper & back end for Isilon.leankit.com website 
API documentation is at Isilon.leankit.com/io 

I have been half successfully package my code:
pip install -e /folder name/ work but pip install . when in the folder does not work

on my workstation at work I have a cron that record the state of my board 2 time a day (cron_job.py)
These data are imported by time_import.py and process by history.py 

if you want an instant report timereport.py is the way to go 

I have not very diligent on commenting my code ... working on it.

I will do a website out of these data currently I generate pretty table and excel sheet (provided in the output folder)

to run it you need an input.json file in the folder you are running the  python script. 
I have created new.py to run the 3 features for you (cron job, history report, instant report)

I will provide input.json to you in an email.

Once again thank you for your help you have made these 2 quarters very interesting!

Thank you !

G.


students/guillaume/quarter_2_project/leankitbackend/leankitbackend/leankitcards.py
```diff
@@ -0,0 +1,297 @@
...
+
+
+class Leankit_Cards:
+    def __init__(self, cards_list=None, date=None):
+        self.date = date
+        if cards_list is not None:
+            self.append(cards_list, date)
+
+    def append(self, cards_list, date=None):
+        if isinstance(cards_list, list):
```
PythonCHB, line 28: what else might it be? and if it is something else, what should happen?

I think you can simply assume it's a sequence, and loop through it.

maybe catch the exception if that doesn't work.

students/guillaume/quarter_2_project/leankitbackend/leankitbackend/dataprocess.py
```diff
@@ -0,0 +1,311 @@
...
+                        'Finished As Planned']
+
+
+def assigned(card, attr='assignedUsers'):
+    name_str = ', '.join([assigned['fullName']
+                          for assigned in getattr(card, attr)])
+    return name_str
+
+
+def created_by(card):
```
PythonCHB, line 24: why the wrapper, rather than having a method or property on the card class that does this?

students/guillaume/quarter_2_project/leankitbackend/leankitbackend/dataprocess.py
```diff
@@ -0,0 +1,311 @@
...
+    table = create_table(data, fields)
+    ret_lst.insert(0, fields)
+
+    return ret_lst, table
+
+
+def table_new_requests(cards, cutoff_date):
+    cards = new_requests(cards, cutoff_date)
+
+    def lst_card(card):
```
PythonCHB, line 241: is there a need for this closure? or could this be defined outside this function or as a method on the card class?

students/guillaume/quarter_2_project/leankitbackend/leankitbackend/dataprocess.py
```diff
@@ -0,0 +1,311 @@
...
+    table = PrettyTable()
+    table.field_names = header
+    for lst in iter:
+        table.add_row(lst)
+    return table
+
+
+def table_cards_per_users(users_cards, users_lst=None):
+
+    def c_data(card):
```
PythonCHB, line 264: same as above -- I don't think this needs to be a closure, and it's not very efficient to re-define the function each time it's used.

students/guillaume/quarter_2_project/leankitbackend/leankitbackend/history.py
```diff
@@ -0,0 +1,150 @@
...
+                 # cvt_d(item.movedOn),
+                 item.plannedFinish]
+            tmp.append(a)
+        return tmp
+
+    def g(value):
+        tmp = []
+        for item in value:
+            tmp.append(item[1:])
+            tmp.sort(key=lambda x: h(x))
```
PythonCHB, line 58: do you want to sort inside the loop? rather than after adding everything?

and you can jsut pass h in for the key function.

and I think this could be a list comp:

```
return sorted([item[1:] for item in value], key=h)
```

students/guillaume/quarter_2_project/leankitbackend/leankitbackend/history.py
```diff
@@ -0,0 +1,150 @@
...
+        return tmp
+
+    def h(lst):
```
PythonCHB, line 61: careful with one letter function names -- you may forget what they mean next time you look at the code months from now...

students/guillaume/quarter_2_project/leankitbackend/leankitbackend/history.py
```diff
@@ -0,0 +1,150 @@
...
+        '''
+        Sorted per year, then per week number using a set
```
PythonCHB, line 63: where is the set here???

students/guillaume/quarter_2_project/leankitbackend/leankitbackend/history.py
```diff
@@ -0,0 +1,150 @@
...
+        '''
+        tmp = lst[0].split('-')
+        tmp = map(int, tmp)
```
PythonCHB, line 66: I'd be inclined to use a comprehension here:

`tuple(int(i) for i in lst[0].split('-')`

students/guillaume/quarter_2_project/leankitbackend/leankitbackend/history.py
```diff
@@ -0,0 +1,150 @@
...
+    dic = {key: f(value) for key, value in projs.items()}
+    ddict = defaultdict(list)
+
+    for key, value in dic.items():
+        for item in value:
+            ddict[item[0]].append([item[1], key, *item[2:]])
+
+    # print(ddict)
+    lst = [[key, value[0][0], *g(value)] for key, value in ddict.items()]
+    lst.sort(key=lambda x: x[1])
```
PythonCHB, line 94: this could be:

`lst.sort(key=lambda x: x[1], reverse=True)`

to save a line of code :-)

students/guillaume/quarter_2_project/leankitbackend/leankitbackend/history.py
```diff
@@ -0,0 +1,150 @@
...
+    format_xl(wb, ws)
+    '''
+    return lst, ret_lst, table
+
+
+def time_report():
+
+    Leankit_input = init()
+    if Leankit_input != {}:
+        global opath
```
PythonCHB, line 130: does this really need to be global? maybe this could all be in a class?

students/guillaume/quarter_2_project/leankitbackend/leankitbackend/leankitapi.py
```diff
@@ -0,0 +1,259 @@
...
+
+# The following functions will abstract the API calls:
+
+# 1. API calls for the cards
+
+def Cards_list(token, **args):
+    '''
+    Abstract API call card:list
+
+    Due to the API Limitation, it can only publish 500 records at once
```
PythonCHB, line 109: what a pain!

students/guillaume/quarter_2_project/leankitbackend/leankitbackend/leankitapi.py
```diff
@@ -0,0 +1,259 @@
...
+        headers = {'Authorization': Bearer}
+        response = get(url, headers=headers)
+        if response.status_code == 200:
+            try:
+                return response.json()
+            except json.decoder.JSONDecodeError:
+                return {}
+
+
+def Cards_info(token, **args):
```
PythonCHB, line 170: seeing token get passed around makes me think you might want to put all this in a class, then it would be self.token...

though if that's the only shared data, maybe point.

students/guillaume/quarter_2_project/leankitbackend/leankitbackend/leankitapi.py
```diff
@@ -0,0 +1,259 @@
...
+                                                 cards=card_dict, **args))
+        loop.run_until_complete(future)
+        tmp = [a.result() for a in future.result()]
+        return [merg_d(info, card) for card, info
+                in zip(tmp, card_dict)]
+    print('there is a pb')
+    return []
+
+
+async def get_calls(token, **args):
```
PythonCHB, line 200: nice use of async!

students/guillaume/quarter_2_project/leankitbackend/leankitbackend/leankitcards.py
```diff
@@ -0,0 +1,297 @@
...
+            self.date = date
+            self.cards_list = []
```
PythonCHB, line 30: I think a single list of card objects, or a card dict would be easier to keep in sync.

students/guillaume/quarter_2_project/leankitbackend/leankitbackend/leankitcards.py
```diff
@@ -0,0 +1,297 @@
...
+                self.status = True
+                # self.json = [card.json for card in self.cards_list]
+            except KeyError:
+                self.status = False
+
+    def projects(self):
+        return projects(self)
+
+    def sorted_projects(self):
+        return sorted_projects(self)
```
PythonCHB, line 53: why is this defined externally?

students/guillaume/quarter_2_project/leankitbackend/leankitbackend/leankitcards.py
```diff
@@ -0,0 +1,297 @@
...
+    def in_progress(self):
+        return self.sorted_projects()['In Progress']
+
+    @property
+    def date(self):
+        return self._date
+
+    @date.setter
+    def date(self, datetime_obj):
+        if isinstance(datetime_obj, datetime):
```
PythonCHB, line 64: what if it's not?

students/guillaume/quarter_2_project/leankitbackend/leankitbackend/leankitcards.py
```diff
@@ -0,0 +1,297 @@
...
+
+    def __repr__(self):
+        if hasattr(self, 'cards_list'):
+            return ('').join([card.__repr__() for card in self.cards_list])
+        return 'There is no cards being imported'
+
+
+class Leankit_Card:
+    def __init__(self, card):
+        if isinstance(card, dict):
```
PythonCHB, line 77: Just use it like a dict, and catch the error if it doesn't work.

When you write code like this, it now won't work with a dict-like object, even though it probably should.

students/guillaume/quarter_2_project/leankitbackend/leankitbackend/leankitcards.py
```diff
@@ -0,0 +1,297 @@
...
+
+    def __repr__(self):
+        try:
+            Name = ''
+            if len(self.assignedUsers) > 0:
+                tmp_lst = [user['fullName'] for user in self.assignedUsers]
+                Name = (', ').join(tmp_lst)
+            tmp_lst = self.type['title'], self.id, self.title, Name
+            return '{} {} {} {}\n'.format(*tmp_lst)
+        except (AttributeError, KeyError):
```
PythonCHB, line 93: repr works on self -- so this is probably NOT where you should be catching that error!

students/guillaume/quarter_2_project/leankitbackend/leankitbackend/leankitcards.py
```diff
@@ -0,0 +1,297 @@
...
+            return 'input was not a card'
+
+
+class Leankit_Board:
+    def __init__(self, board):
+        if isinstance(board, dict):
```
PythonCHB, line 99: you probably don't need to check for a type at all -- but if you do, you want to use:

`if instance(collections.abc.Mapping): `

We didn't get into Abstract Base Classes in class, as they aren't very widely used, but it's a way to check that you've got a "mapping" type, which would be a dict, but also any other type that acts like a dict.

students/guillaume/quarter_2_project/leankitbackend/leankitbackend/leankitcards.py
```diff
@@ -0,0 +1,297 @@
...
+lanes_Backlogs_next_sprint = ['Backlog', 'Next Sprint Planning']
+
+
+# Attention sub columns of recently finished / ready to archived has the
+# same name
+lanes_Finished = ['Finished As Planned', 'Started but not Finished',
+                  'Discarded or Obsolete']
+
+
+class Leankit_Cards:
```
PythonCHB, line 21: just a style issue -- usual use either capitalization or underscores, but not both:

`LeankitCards`

students/guillaume/quarter_2_project/leankitbackend/leankitbackend/leankitcards.py
```diff
@@ -0,0 +1,297 @@
...
+            return [lane['name'], lane['id'], lane['index']]
+        try:
+            lst_1 = [la(lane) for lane in self.lanes
+                     if lane['parentLaneId'] is None]
+            d = defaultdict(list)
+            for lane in self.lanes:
+                if lane['parentLaneId'] is not None:
+                    d[lane['parentLaneId']].append(la(lane))
+
+            def g(key):
```
PythonCHB, line 148: could some of these functions that unpack data structures be methods of a class instead? that way, only the class needs to know how the data is structured.

students/guillaume/quarter_2_project/leankitbackend/leankitbackend/leankitcards.py
```diff
@@ -0,0 +1,297 @@
...
+
+            return [lst_1, e]
+
+        except (KeyError, AttributeError):
+            return []
+
+
+def cards_per_types(cards, attribute='type', key='title'):
+    dic = defaultdict(list)
+    if isinstance(cards, Leankit_Cards):
```
PythonCHB, line 165: if you made Leankit_Cards an iterable, then this could jsut be:

`cards_lst - list(cards)`

with no type checking or anything.

students/guillaume/quarter_2_project/leankitbackend/leankitbackend/util.py
```diff
@@ -0,0 +1,290 @@
...
+
+def printd(dic):
+    '''
+    Print a dictionnary in a more readable way
+    '''
+    for key, val in dic.items():
+        print('{} {}'.format(key, val))
+
+
+def printl(lst):
```
PythonCHB, line 23: check out the prettyprint module

```
from pprint import pprint

pprint(something)
```
saves you having to write this yourself...

students/guillaume/quarter_2_project/leankitbackend/setup.py
```diff
@@ -0,0 +1,15 @@
...
+
+setup(name='leankitbackend',
+      install_requires=['PTable>=0.8', 'xlsxwriter', 'asyncio',
+                        'request', 'requests', 'aiohttp'],
+      version='1.0',
+      description='Python Distribution Utilities',
+      author='Guillaume THOMAS',
+      author_email='guillaume.thomas@dell.com',
+      url='https://github.west.isilon.com/eng-tools/Lenkit_Backend',
+      # packages=find_packages()
```
PythonCHB, line 14: This may be why your setup.py doesn't completely work -- you need to define the packages to install one way or another. either by hand or with find_packages().

PythonCHB (approved):
really nice work -- great job making us onf a API and modeling your data.

Thanks for your insightful questions during class.

Enjoy the web dev class!

PR #197, opened 2018-03-16
submission for final class project

mattmaeda:
Here are the test results.  Needed to use the built in test framework for Django to run my tests.

(venv) 15:18:24[matt@luna:stockscreener](master)$django-admin test exchange
Creating test database for alias 'default'...
System check identified no issues (0 silenced).
..................................
----------------------------------------------------------------------
Ran 34 tests in 0.154s

OK
Destroying test database for alias 'default'...


students/mattmaeda/stockscreener/exchange/views.py
```diff
@@ -0,0 +1,266 @@
...
+            ticker = request.POST.get("ticker", "")
+            symbol = ticker.upper()
+            company = get_stock_info(symbol)
+
+            if company is not None:
+                history = get_price_history(symbol)
+                response = run_screens(history)
+                message = "Here are your results for {}".format(symbol)
+            else:
+                message = "Invalid stock symbol '{}'".format(symbol)
```
codedragon, line 31: There is a lot of code in this try block. Think about where you expect to get the ConnectionError, and try to limit the block in the try block to just that. Especially try to avoid conditionals in a try block.

students/mattmaeda/stockscreener/exchange/views.py
```diff
@@ -0,0 +1,266 @@
...
+    try:
+        stock = Stock(ticker.upper())
+        return stock.get_company().get("companyName")
+
+    except iexfinance.utils.exceptions.IEXSymbolError as exc:
+        return None
+
+    except ValueError as ve:
+        # If empty
+        return None
```
codedragon, line 60: You should at least log the exceptions you are catching. And are you sure that you don't want the user to know about them at all?

codedragon (approved):
Really nice job, Matt. Tests, doc strings, and packaged up. And you are ahead of the game for next quarter with web frameworks.

PR #198, opened 2018-03-17
Final commit for the quarter

danwoj:
I pushed a bunch of demos and POCs but my final project is dlpscan.py as well as its test code test_dlpscan.py. Admittedly, I didn't get all the testing in that I should have.


students/danwoj/dlpscan/dlpscan.py
```diff
@@ -0,0 +1,322 @@
...
+would have taken a lot more time to implement it here '''
+
+#!/usr/bin/env python
+import winrm, os, socket, csv, nacl.pwhash
+import subprocess as sub
+from pexpect import pxssh
+
+s = pxssh.pxssh(timeout=60)
+password = 'L!ghtning26'
+account = 'dlpscan.svc'
```
codedragon, line 17: You definitely do not want to have passwords on any files that you are uploading to github. Use a configuration file that is not in your git repo, environmental variables or a secret keeper (like Vault).

students/danwoj/dlpscan/dlpscan.py
```diff
@@ -0,0 +1,322 @@
...
+            else:
+                bin_subnet_address.append('0')
+                octets[i] += '0'
+            dec_octets[i] = int(octets[i], 2)
+        final_subnet_address += str(dec_octets[i])
+        if i < 3:
+            final_subnet_address += '.'
+    final_subnet_address += '/'
+    final_subnet_address += str(cidr)
+    return final_subnet_address
```
codedragon, line 70: Did you look at iptools? http://python-iptools.readthedocs.io/en/latest/

students/danwoj/dlpscan/test_dlpscan.py
```diff
@@ -0,0 +1,31 @@
...
+    subnet_mask2 = '255.255.248.0'
+    value2 = subnet_calculator(ip_addr2, subnet_mask2)
+
+    ip_addr3 = '172.20.23.128'
+    subnet_mask3 = '255.255.0.0'
+    value3 = subnet_calculator(ip_addr3, subnet_mask3)
+
+    assert value1 == '192.168.36.0/24'
+    assert value2 == '10.2.32.0/21'
+    assert value3 == '172.20.0.0/16'
```
codedragon, line 21: These should be three separate tests, so if one fails, you can easily tell which one failed, and the others will still run, so you will know if it is just one failure or three. You can use parameterize (or pytests version of parameterize) to do this pretty easily.

students/danwoj/dlpscan/dlpscan.py
```diff
@@ -0,0 +1,322 @@
...
+        return subnet
+
+    def gateway(self):
+        default_gate = sub.Popen("ip route show | grep default | awk -F ' ' '{print $3}'", stdout = sub.PIPE, shell = True)
+        (output, err) = default_gate.communicate()
+        default_gate = (str(output, "utf8").split("\n")).pop(0)
+
+        return default_gate
+
+def sudo(s,password):
```
codedragon, line 115: I would not call this function sudo. While not a python keyword, your program is doing enough terminal commands, that I would be worried about users (or other developers) becoming confused about its purpose.

students/danwoj/dlpscan/dlpscan.py
```diff
@@ -0,0 +1,322 @@
...
+quite resolve the Windows RM issue. Had I had more time, I was also experimenting with PyNaCl to appropriate input and store my credentials 
+I'd use to perform the remote login and save these passwords in an encrypted 'shadow' file. I found some success with this in iPython but it 
+would have taken a lot more time to implement it here '''
+
+#!/usr/bin/env python
+import winrm, os, socket, csv, nacl.pwhash
+import subprocess as sub
+from pexpect import pxssh
+
+s = pxssh.pxssh(timeout=60)
```
codedragon, line 15: Please call this something besides s. Not at all clear what it is or where it comes from. It really would be better to not have it as a global, as well, since that makes it really difficult to keep track of its state.

codedragon (approved):
Nice job, you are making progress. Estimating out how long something is going to take is one of the most difficult parts of writing code, and we often underestimate. As you surmised, I would like to see more tests. :) But, you have a nice start here.

codedragon:
Oh, and please add doc strings to your functions and classes, makes it much easier to tell what's going on.

PR #199, opened 2018-03-17
Personal project updates

chops77:


students/Chao/project_fpy/fpy.py
```diff
@@ -0,0 +1,253 @@
...
+class Pool(LTM):
+    """ Pool class for retrieving pool status """
+
+    def get_pool_state(self, p_name, p_partition='Common'):
+        """ Function to retrieve current pool availability status """
+        pool_stat = self.mgmt.tm.ltm.pools.pool.load(name=p_name, partition=p_partition)
+        stats = Stats(pool_stat.stats.load())
+        return stats.stat.status_availabilityState['description']
+
+    def get_pool(self):
```
codedragon, line 36: Since you aren't actually printing anything, better to call this print_pool_status or something like that.

students/Chao/project_fpy/fpy.py
```diff
@@ -0,0 +1,253 @@
...
+            rule_path = '/{}/{}'.format(virtual.partition, myrule)
+            vrule = self.mgmt.tm.ltm.virtuals.virtual.load(name=virtual.name, partition=virtual.partition)
+            try:
+                if rule_path in vrule.rules:
+                    print("Rule {} already exists in {}".format(myrule, virtual.name))
+                else:
+                    # Insert iRule at the front of the list
+                    vrule.rules.insert(0, rule_path)
+                    vrule.update()
+                    print("Rule {} added to {}".format(myrule, virtual.name))
```
codedragon, line 79: I think this would be better:
```
if rule_path in vrule.rules:
    print("Rule {} already exists in {}".format(myrule, virtual.name))```
    return
try:
    # Insert iRule at the front of the list
    vrule.rules.insert(0, rule_path)
    vrule.update()
except iControlUnexpectedHTTPError:
    print("Unable to apply rule {} to {}. HTTP profile required.".format(myrule, virtual.name))
else:
    print("Rule {} added to {}".format(myrule, virtual.name))
```
It is pretty minor, really, but having conditionals in try/except blocks can lead to confusion and occasionally unexpected errors.

students/Chao/project_fpy/fpy.py
```diff
@@ -0,0 +1,253 @@
...
+        result = input("Please select a menu item: ")
+
+        if result == '1':
+            r = Rule(mgmt)
+            rule_name = input("Please enter iRule name (optioanl): ")
+            if rule_name:
+                r.create_rule(rule_name)
+            else:
+                r.create_rule()
+        elif result == '2':
```
codedragon, line 174: Conditionals 2 and 3 are getting large enough, I would put them in their own function. Actually, they are both the same, except for the words apply vs remove, so you could use one function that flips between apply and remove, either by a dictionary or a conditional.

codedragon (approved):
Great job. Some minor comments, but overall well organized and pythonic code. Would be great to see it packaged, as well. Very happy to see tests. :)

PR #200, opened 2018-03-18
class 2 project

PythonCHB:
This was emailed to Chris and Maria, and added to the repo by Chris for review.


PythonCHB:
@geneuv: I've put this in your dir in the repo for review -- follow along!

PythonCHB:
First:

Ideally this would be in a package.

But even if not, a little README is a good idea, so we (and anyone else looking at the code) can see what it is supposed to do and how to use it.

students/gene/project/node_db.py
```diff
@@ -0,0 +1,131 @@
...
+        self.management_ip = management_ip
+        self.console_id = console_id
+        self.vendor_id = vendor_id
+        self.platform_id = platform_id
+        self.os_id = os_id
+
+
+    @property
+    def description(self):
+        return ("""Node: {}
```
PythonCHB, line 20: the built in textwrap.dedent() function is handy for this sort or thing:

https://docs.python.org/3/library/textwrap.html?highlight=dedent#textwrap.dedent

students/gene/project/node_db.py
```diff
@@ -0,0 +1,131 @@
...
+Site: {}
+Management IP: {}
+Console: {}
+Vendor: {}
+Platform: {}
+SW Version: {}""".format(self.node_id, self.site_id, self.management_ip,
```
PythonCHB, line 26: this would be a great place to use "f strings":

f"""Site: {self.site_id}"""

etc -- that is, you can put simple expression right in the brackets

students/gene/project/node_db.py
```diff
@@ -0,0 +1,131 @@
+
+import sqlite3
+from sqlite3 import Error
+
+
+class Node:
```
PythonCHB, line 6: at this point, a Node object doesn't have any logic -- so maybe it would be better to simply use a dict to store all that -- less boilerplate. Though if you forsee logic being added, then yes, a class is the right way to go.

students/gene/project/node_db.py
```diff
@@ -0,0 +1,131 @@
...
+class NodeStorage:
+    """
+    To test in database in memory use ':memory:'
+    To use database use 'test.db'
+
+    """
+
+    def __init__(self, database='test.db'): 
+        self.db = database
+        self.conn = sqlite3.connect(self.db)
```
PythonCHB, line 40: maybe this should call self.create_connection() instead? Better not to have the same code in more than one place.

students/gene/project/node_db.py
```diff
@@ -0,0 +1,131 @@
...
+        self.create_table()
+
+
+    def create_connection(self):
+        try:
+            conn = sqlite3.connect(self.db)
+            return conn
+        except Error as e:
+            print(e)
```
PythonCHB, line 49: this is going to print the error, but swallow it -- do you want the program to keep running if it can't make a connection?

students/gene/project/node_db.py
```diff
@@ -0,0 +1,131 @@
...
+                      {'node_id': node_info.node_id, 'site_id': node_info.site_id, 'management_ip': node_info.management_ip,
+                       'console_id': node_info.console_id, 'vendor_id': node_info.vendor_id, 'platform_id': node_info.platform_id,
+                       'os_id': node_info.os_id})
+
+
+    def remove_node(self, node_id):
+        with self.conn:
+            c = self.conn.cursor()
+            c.execute("DELETE FROM node_db WHERE node_id = :node_id", {'node_id': node_id})
+            return node_id
```
PythonCHB, line 89: what are you going to get if you try to delete a node that doesn't exist?

students/gene/project/node_db.py
```diff
@@ -0,0 +1,131 @@
...
+    def find_node(self, node_id):
+        with self.conn:
+            c = self.conn.cursor()
+            c.execute("SELECT * FROM node_db WHERE node_id = :node_id", {'node_id': node_id})
+            node = c.fetchone()
+
+        if node:
+            return Node(*node)
+        else:
+            return None
```
PythonCHB, line 101: what is node if it isn't found? is it already None?

students/gene/project/node_db.py
```diff
@@ -0,0 +1,131 @@
...
+            for node in all_devices:
+                # print(donor)
+                node_id = node[0]
+                site_id = node[1]
+                management_ip = node[2]
+                vendor_id = node[3]
+                platform_id = node[4]
+                os_id = node[5]
+
+                print("{:12}{:7}{:16}{:10}{:11}{:11}".format(node_id, site_id, management_ip, vendor_id, platform_id, os_id))
```
PythonCHB, line 123: as you had to extract them all from node by index, why not pass `*node` into `format()` instead?

students/gene/project/test_node_db.py
```diff
@@ -0,0 +1,38 @@
+from node_db import Node, NodeStorage
+import pytest
+
+
+
+
+@pytest.fixture
+def sample_data():
+
+    ds = NodeStorage()
```
PythonCHB, line 10: I would pass :memory: directly in here -- as a fixture you want to be real sure you're making a clean DB that won't mess anything else up.

students/gene/project/test_node_db.py
```diff
@@ -0,0 +1,38 @@
+from node_db import Node, NodeStorage
+import pytest
+
+
+
+
+@pytest.fixture
```
PythonCHB, line 7: great use of a fixture!

students/gene/project/test_node_db.py
```diff
@@ -0,0 +1,38 @@
...
+def test_find_node_notthere(sample_data):
+    device = sample_data.find_node('TGRIRV30')
+    assert device is None
+
+
+def test_remove_node(sample_data):
+    device = sample_data.remove_node('TRNSOY30')
+    device2 = sample_data.find_node('TRNSOY30')
+    assert device2 is None
+
```
PythonCHB, line 38: great start on testing -- but you peobably want to add some more -- is the node object retrieved from teh DB have all the values it should have?

here's where adding an __eq__ to your Node class could help, then you can make sure the retrieved-from-the-db Node matches the original.

students/gene/project/node_app.py
```diff
@@ -0,0 +1,89 @@
+
+
+from node_db import Node, NodeStorage
+
+
+def main_menu_usage():
```
PythonCHB, line 6: I'd probably just make this a big string, rather than a function. But not that different

students/gene/project/node_app.py
```diff
@@ -0,0 +1,89 @@
...
+
+def main_menu():
+
+    while True:
+        main_menu_usage()
+        try:
+            answer = int(input("Option: "))
+            print("\nOption selected: {}\n\n".format(answer))
+        except ValueError:
+            answer = 0
```
PythonCHB, line 26: no need to set answer here -- it will get reset next time through the loop.

students/gene/project/node_app.py
```diff
@@ -0,0 +1,89 @@
...
+            print("\n\n****ERROR: Please choose a numeric option****\n\n")
+            continue
+
+        if answer == 4:
+            print("\nQuiting.....")
+            break
+
+        elif answer == 1:
+            node_id = input("Device name: ").strip().upper()
```
PythonCHB, line 35: this logic should all be in a separate function, called from in the interactive loop.

students/gene/project/node_app.py
```diff
@@ -0,0 +1,89 @@
...
+            site_id = input("Site: ").strip().upper()
+            management_ip = input("Management IP: ").strip()
+            console_id = input("Console server: ").strip().upper()
+            vendor_id = input("Vendor: ").strip().upper()
+            platform_id = input("Platform: ").strip().upper()
+            os_id = input("SW Version: ").strip()
```
PythonCHB, line 41: probably a good idea to have some validation of the input at some point....

students/gene/project/node_app.py
PythonCHB, line 35: this is a lot of repeated code -- can you use setattr() and a loop to make it less repeated?

and/or store it all ina dict, then create the Node object from a dict?

IT would look somethign like:
```
attributes = ["site_id", "console_id', ...]
for at in attributes:
    atts[at] = input(f"please type in: {at}:").strip()

```

students/gene/project/node_app.py
```diff
@@ -0,0 +1,89 @@
...
+                    print("\n\nDevice has been added to database\n\n")
+                else:
+                    continue
+
+            else:
+                print("\n\nDevice {} already exist in databse with below information\n\n".format(device.node_id))
+                print("--------------------------")
+                print(device.description)
+                print("--------------------------")
+            # print("\nAdding Device to Database...\n\n\n")
```
PythonCHB, line 63: shouldn't there be a `continue` here?

PythonCHB (commented):
Basic, but it looks useful.

I do encourage you to take the time to really clean this up and flesh it out -- so you can get the hang of making a package, comprehensive testing, etc.

Enjoy the third quarter!

PR #201, opened 2018-03-19
Final Project

zandra2:
Submitting Final Project.


students/zandra_eng/finalproject/test_web.py
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env python 3
```
PythonCHB, line 1: you actually need "python3" without the space for this to work :-)

students/zandra_eng/finalproject/web.py
```diff
@@ -0,0 +1,66 @@
...
+    if len(sys.argv) < 4 :
+        print("Error: Give a stock name, start and end dates as arguments")
+        exit(1)
+    stock_name = sys.argv[1]
+    start = parse(sys.argv[2])
+    end = parse(sys.argv[3])
+    plotter = StockPlotter()
+    stocks = plotter.fetch_stock(stock_name, start, end)
+    csvfile = "{}.csv".format(stock_name)
+    plotter.save(stocks, csvfile)  #./ = current folder
```
PythonCHB, line 64: if you're going to put this in a class, (which is a good idea), then you should save thigs like the stock_name in the class when you load it -- then you do'nt need to keep passing it around:

```
plotter.StockPlotter()
plotter.fetch_stock(stock_name, start, end)
plotter.save()
plotter.plot()
```
see my notes in the code as to how to make that work.

students/zandra_eng/finalproject/web.py
```diff
@@ -0,0 +1,66 @@
...
+# if pulling data from the internet for python 2 use: import pandas.io.data as web
+
+from pandas_datareader import data as web
+
+
+class StockPlotter:
+
+    def fetch_stock(self, stock_name, start, end):
+        df = web.DataReader(stock_name, "google", start, end)
+        return df
```
PythonCHB, line 30: save these parameters in the instance (self):

```
self.name = stock_name
self.start = start # you maybe don't need these, they are in the data, yes?
self.end = end 
self.df = web.DataReader(stock_name, "google", start, end)
```
now you have them for the other methods to work with.

students/zandra_eng/finalproject/web.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def save(self, stocks, filename):
```
PythonCHB, line 32: so this can be:

```
def save(self):
    filename = self.name + ".csv"
    self.df.so_csv(filename)
```

students/zandra_eng/finalproject/web.py
```diff
@@ -0,0 +1,66 @@
...
+        stocks.to_csv(filename)
+
+    def plot(self, stock_name, csvfile):
+        dates = []
+        values = []
+        with open(csvfile) as csvfile:
```
PythonCHB, line 38: Couldn't you plot the DataFrame directly? might as well re-use it if you already created it.

Then have another method for loading up a dataframe from a csv file instead of getting it from the web.

students/zandra_eng/finalproject/web.py
```diff
@@ -0,0 +1,66 @@
...
+            reader = csv.reader(csvfile, delimiter=',')
```
PythonCHB, line 39: pandas has a nice csv reader built in :-)

students/zandra_eng/finalproject/test_web.py
```diff
@@ -0,0 +1,33 @@
...
+from web import StockPlotter
+import csv
+expected = [[88.14,  90.50,  88.07,  89.98,  13383214],
+            [90.35,  90.50,  88.63,  89.06,  11054682],
+            [87.98,  88.53,  87.34,  87.74,  11453034]]
+
+def test_fetch_stock():
+
+    s = StockPlotter()
+    stocks = s.fetch_stock('WMT', '3/5/2018', '3/7/2018')
```
PythonCHB, line 13: This is where you really want mocking -- this test will only run if the web service is available -- and hopefully fast!

But still good to have something!

students/zandra_eng/finalproject/test_web.py
```diff
@@ -0,0 +1,33 @@
...
+        for cindex, col in enumerate(row):
+            assert col == stocks.ix[rindex, cindex]
+
+def test_save_stock():
+
+    s = StockPlotter()
+    stocks = s.fetch_stock('WMT', '3/5/2018', '3/7/2018')
+    s.save(stocks, 'test.csv')
+    with open('test.csv') as csvfile:
+        reader = csv.reader(csvfile, delimiter=',')
```
PythonCHB, line 25: this looks exactly like the reading code in your plotting -- so better have a csv reading funciton in your class, then you can use it in both places.

DRY (Don't Repeat Yourself) -- always good guidance for coding!

PythonCHB (approved):
Nice job figuring out a bunch of libraries!

Do take some time to think about the structure -- for something this small, it doesn't much matter, but now is the time to get good habits.

PythonCHB:
Thanks for being such a positive presence in class and office hours.

Keep coding, the only way to learn to code is to write code.

Good luck with your future endeavours,

-Chris

